PR #2, opened 2018-10-02
The plugins section is optional

jwendell:
If '.plugins[]' is not found in original file, jq fails to patch it.


tiswanso:
nice!  I'll test it and if it works, merge it.

Issue #4, opened 2018-10-04
Create Istio-cni helm chart


tiswanso:
Create a helm chart for templating the istio-cni install manifest.

PR #9, opened 2018-10-04
Add helm chart

tiswanso:
Create the helm chart for the istio-cni deployment with default values.yaml and GKE settings in values_gke.yaml.  

Closes: #4

### DIFF
--- a/deployments/kubernetes/install/helm/values.yaml
+++ b/deployments/kubernetes/install/helm/values.yaml
@@ -1,0 +1,20 @@
+hub: docker.io/tiswanso
+tag: v0.1-dev
+pullPolicy: Always
+
+logLevel: info
+
+# Configuration file to insert istio-cni plugin configuration
+# by default this will be the first file found in the cni-conf-dir
+# Example
+# cniConfFileName: 10-calico.conflist
+
+# CNI bin and conf dir override settings
+# defaults:
+cniBinDir: /opt/cni/bin
+cniConfDir: /etc/cni/net.d
+
+excludeNamespaces:
+  - istio-system
+
+
--- a/deployments/kubernetes/install/helm/.helmignore
+++ b/deployments/kubernetes/install/helm/.helmignore
@@ -1,0 +1,1 @@
+NOTES.txt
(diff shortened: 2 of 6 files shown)


deployments/kubernetes/install/scripts/install-cni.sh
```diff
@@ -141,7 +141,7 @@ sed -i s~__KUBECONFIG_FILEPATH__~${HOST_CNI_NET_DIR}/istio-cni-kubeconfig~g $TMP
 sed -i s~__LOG_LEVEL__~${LOG_LEVEL:-warn}~g $TMP_CONF
 
 # default to first file in `ls` output
-CNI_CONF_NAME=${CNI_CONF_NAME:-$(ls | head -n 1)}
+CNI_CONF_NAME=${CNI_CONF_NAME:-$(ls /host/etc/cni/net.d | head -n 1)}
```
jwendell: While I think it's good (I was using that too) I think it's better to keep this as a separate PR.
Also, note that at this point the file `istio-cni-kubeconfig` is already present on that dir, and it might end up being the first file :)

jwendell: In minikube they're not using a numerical pattern, so, `istio-cni-kubeconfig` comes first. What about putting that file in another directory?

tiswanso: Ahh... ok.  The problem with another dir is that it has to be a host dir that we'd have to mount (via k8s Volume).  The creation of the k8s conf file in that dir is following the calico pattern.  I think for now popping some number of `Z`s on the front is probably the safest & simplest approach, ie. `ZZZ-istio-cni-kubeconfig`.  I don't love that but kubelet doesn't have a config for a specific CNI conf file to use, just the dir.

deployments/kubernetes/install/helm/values.yaml
```diff
@@ -0,0 +1,20 @@
+hub: docker.io/tiswanso
```
john-a-joyce: Is this right?

tiswanso: for now, yes... I've built and pushed the images there and that's where folks have been using them from.  I don't think I have access to push anywhere else.  Another proj thing to get working when CI is integrated.

Issue #3, opened 2018-10-04
Config insertion not compatible with singular CNI plugin config (no .plugins[])


tiswanso:
If the existing CNI plugin configuration has a top level dictionary/map that doesn't have a .plugins array then the istio-cni config insertion will not be effective since it will add the .plugins array under the existing single plugin's config.

Example:
Existing plugin config: 
```
{
  "name": "rkt.kubernetes.io",
  "type": "bridge",
  "bridge": "mybridge",
  "mtu": 1460,
  "addIf": "true",
  "isGateway": true,
  "ipMasq": true,
  "ipam": {
    "type": "host-local",
    "subnet": "10.1.0.0/16",
    "gateway": "10.1.0.1",
    "routes": [
      {
        "dst": "0.0.0.0/0"
      }
    ]
  }
}
```

Insertion result:
```
{
  "name": "rkt.kubernetes.io",
  "type": "bridge",
  "bridge": "mybridge",
  "mtu": 1460,
  "addIf": "true",
  "isGateway": true,
  "ipMasq": true,
  "ipam": {
    "type": "host-local",
    "subnet": "10.1.0.0/16",
    "gateway": "10.1.0.1",
    "routes": [
      {
        "dst": "0.0.0.0/0"
      }
    ]
  },
  "plugins": [
    {
      "type": "istio-cni",
      "log_level": "info",
      "kubernetes": {
        "kubeconfig": "/etc/cni/net.d/istio-cni-kubeconfig",
        "cni_bin_dir": "/opt/cni/bin",
        "exclude_namespaces": [
          "istio-system"
        ]
      }
    }
  ]
}
```

tiswanso:
Thoughts on the fix:
if   .type exists at top level,  grab the whole dict and insert it into the plugins section of a new dict of the form:
```
{
  "name": "k8s-pod-network",
  "cniVersion": "0.3.0",
  "plugins": [
   ]
}
```

jwendell:
I'm working on this

tiswanso:
@jwendell  -- did you want to handle the filename thing under here?  or with a new issue?

PR #11, opened 2018-10-05
Fix CNI config insertion on files that don't contain .plugins[]

jwendell:
Closes: #3.

### DIFF
--- a/deployments/kubernetes/install/scripts/install-cni.sh
+++ b/deployments/kubernetes/install/scripts/install-cni.sh
@@ -158,14 +158,21 @@ CNI_CONF_FILE=${MOUNTED_CNI_NET_DIR}/${CNI_CONF_NAME}
 if [ -e "${MOUNTED_CNI_NET_DIR}/${CNI_CONF_NAME}" ]; then
     # This section overwrites an existing plugins list entry to for istio-cni
     CNI_TMP_CONF_DATA=$(cat ${TMP_CONF})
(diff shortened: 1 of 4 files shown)


tiswanso:
Looks good...  I'm reviewing and testing.

tiswanso:
@jwendell I ran this on a ubuntu system with plugin conf file with this content:

tiswanso:
It looks like it replaces just the `type` field in the top level dict

jwendell:
@tiswanso thanks for testing it, I've updated the jq filter. Please, try again.
It looks like we need a testsuite though.

deployments/kubernetes/install/scripts/install-cni.sh
```diff
@@ -155,14 +155,21 @@ CNI_CONF_FILE=/host/etc/cni/net.d/${CNI_CONF_NAME}
...
     echo "${CNI_CONF_DATA}" > ${TMP_CONF}
 fi
 
 # Delete old CNI config files for upgrades.
 if [ "${CNI_CONF_NAME}" != "${CNI_OLD_CONF_NAME}" ]; then
     rm -f "/host/etc/cni/net.d/${CNI_OLD_CONF_NAME}"
 fi
+
+# If the old config filename ends with .conf, rename it to .conflist, because it has changed to be a list
+if [ "${CNI_CONF_NAME: -5}" == ".conf" ]; then
```
tiswanso: This section is ending up creating another separate file and not removing the old file.  IMO, we should either just overwrite the old file or remove it and create the .conflist file.  If you put this `if` section prior to the section at line 162 it will `rm` the old file.

tiswanso:
The jq filter seems to be working now.  I checked it for both a file without any `plugins` array and one with an existing `plugins` array.

Issue #12, opened 2018-10-05
Fix detect first file in cni-conf-dir and ensure istio-kubeconfig isn't first


tiswanso:
The install logic doesn't detect the first file in the cni-conf-dir.  It's just grabbing the first file in the default dir.

Also, the install logic should ensure istio-kubeconfig isn't the first file in the cni-conf-dir

PR #14, opened 2018-10-08
Fix first file detection in installation script

venilnoronha:
This PR ensures that the first configuration file is located from within the `/host/etc/cni/net.d/` directory, and that it also ignores any files containing `istio-cni-kubeconfig` in their name.

Fixes #12.

Signed-off-by: Venil Noronha <veniln@vmware.com>

PR #15, opened 2018-10-08
Install: ensure generated kubeconfig file is not the first file in cni-conf-dir

tiswanso:
For now, name the generated kubeconf file "ZZZ-istio-cni-kubeconfig".

Also, fix command to find first file in cni-conf-dir.

Closes: #12

### DIFF
--- a/deployments/kubernetes/install/scripts/install-cni.sh
+++ b/deployments/kubernetes/install/scripts/install-cni.sh
@@ -78,6 +78,8 @@ ${CNI_NETWORK_CONFIG}
 EOF
 fi
 
+KUBECFG_FILE_NM=${KUBECFG_FILE_NM:-ZZZ-istio-cni-kubeconfig}
+
 SERVICE_ACCOUNT_PATH=/var/run/secrets/kubernetes.io/serviceaccount
 KUBE_CA_FILE=${KUBE_CA_FILE:-$SERVICE_ACCOUNT_PATH/ca.crt}
 SKIP_TLS_VERIFY=${SKIP_TLS_VERIFY:-false}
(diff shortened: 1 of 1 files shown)


deployments/kubernetes/install/scripts/install-cni.sh
```diff
@@ -134,14 +136,14 @@ fi
...
+sed -i s/__KUBECONFIG_FILENAME__/${KUBECFG_FILE_NM}/g $TMP_CONF
 
 # Use alternative command character "~", since these include a "/".
-sed -i s~__KUBECONFIG_FILEPATH__~${HOST_CNI_NET_DIR}/istio-cni-kubeconfig~g $TMP_CONF
+sed -i s~__KUBECONFIG_FILEPATH__~${HOST_CNI_NET_DIR}/${KUBECFG_FILE_NM}~g $TMP_CONF
 sed -i s~__LOG_LEVEL__~${LOG_LEVEL:-warn}~g $TMP_CONF
 
 # default to first file in `ls` output
-CNI_CONF_NAME=${CNI_CONF_NAME:-$(ls | head -n 1)}
+CNI_CONF_NAME=${CNI_CONF_NAME:-$(ls /host/etc/cni/net.d | head -n 1)}
```
venilnoronha: I think it should be `ls -1 /host/etc/cni/net.d` as in https://github.com/istio-ecosystem/cni/pull/14/files#diff-a1fbe10058a1077a440d75c2abf5f1e1R144 on PR #14. The `-1` will ensure that the file names are printed one per line.

venilnoronha: Never mind, `ls /host/etc/cni/net.d | head -n 1` will give the same output as `ls -1 /host/etc/cni/net.d | head -n 1)`.

deployments/kubernetes/install/scripts/install-cni.sh
```diff
@@ -78,6 +78,8 @@ ${CNI_NETWORK_CONFIG}
 EOF
 fi
 
+KUBECFG_FILE_NM=${KUBECFG_FILE_NM:-ZZZ-istio-cni-kubeconfig}
```
venilnoronha: nit: I'd prefer `KUBECFG_FILE_NAME` over `KUBECFG_FILE_NM`, just to keep it consistent with the rest of the variable names.

venilnoronha: nit: Can we have a real number prefix, like `99-istio-cni-kubeconfig` or so?

tiswanso: will do

tiswanso: The numerical prefix will always be before the alphabetical prefixes and @jwendell was having problems on minikube where the file is named something that has an alphabetical prefix, e.g. something like "minikube_cni.conf".  I forget the exact name.  So, `ZZZ-` puts it at the lexicographical end for both cases.

venilnoronha: Got it, thanks!

PR #18, opened 2018-10-08
Update check-tree path matching

### LINKED ISSUE #16: Fix GOPATH check in Makefile
Currently the GOPATH check is hardcoded to a specific user relative path (tiswanso/...).  That needs to be changed.  

Probably the path check could be more generic to check a simple regex with `cni` in it.

venilnoronha:
This PR updates `ISTIO_CNI_RELPATH` to `istio.io/cni` in order to enforce a standard location for the Istio CNI plugin.

Fixes #16.

Signed-off-by: Venil Noronha <veniln@vmware.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,8 +15,7 @@
 #-----------------------------------------------------------------------------
 # Global Variables
 #-----------------------------------------------------------------------------
-ISTIO_CNI_RELPATH ?= github.com/tiswanso/istio-cni
-#ISTIO_CNI_RELPATH ?= github.com/tiswanso/istio-cni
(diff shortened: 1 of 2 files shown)


venilnoronha:
@tiswanso I've updated the location to be `istio.io/cni` keeping in mind that this repository will be eventually moved to the `istio` organization. Let me know if you still think that we need a regex match instead.

tiswanso:
Makefile looks good.  Could you update the build section of the README as well?  I guess it should tell folks to clone to a path like `go/src/istio.io/` or to build from their clone location like `ISTIO_CNI_RELPATH=$(pwd) GOOS=linux make build`

Issue #10, opened 2018-10-05
Uninstall mechanism


jwendell:
Should we have a proper uninstall mechanism?

Simply issuing `kubectl delete -f <deploy.yaml>` is not enough. We should revert all changes in the nodes made by `install-cni.sh`.

tiswanso:
I think an option is to have `install-cni.sh` catch sigkill and remove istio-cni from the plugin chain and cleanup files.

PR #20, opened 2018-10-09
istio-cni-node add exit signal handler to clean config

tiswanso:
- Remove istio-cni plugin config from config chain.
- remove kubeconfig file
Closes: #10

### DIFF
--- a/deployments/kubernetes/install/scripts/install-cni.sh
+++ b/deployments/kubernetes/install/scripts/install-cni.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 # Script to install Istio CNI on a Kubernetes host.
 # - Expects the host CNI binary path to be mounted at /host/opt/cni/bin.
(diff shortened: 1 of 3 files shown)


deployments/kubernetes/install/scripts/install-cni.sh
```diff
@@ -9,9 +9,26 @@
 set -u -e
 
 # Capture the usual signals and exit from the script
-trap 'echo "SIGINT received, simply exiting..."; exit 0' SIGINT
-trap 'echo "SIGTERM received, simply exiting..."; exit 0' SIGTERM
-trap 'echo "SIGHUP received, simply exiting..."; exit 0' SIGHUP
+#trap 'echo "SIGINT received, simply exiting..."; exit 0' SIGINT
+#trap 'echo "SIGTERM received, simply exiting..."; exit 0' SIGTERM
+#trap 'echo "SIGHUP received, simply exiting..."; exit 0' SIGHUP
+
```
jwendell: You can get rid of those comments :)

deployments/kubernetes/install/scripts/install-cni.sh
```diff
@@ -9,9 +9,26 @@
...
+function cleanup() {
```
jwendell: For readability, I think it's better to move this function to below the definition of HOST_CNI_NET_DIR and MOUNTED_CNI_NET_DIR variables.

PR #31, opened 2018-10-18
CircleCI job for e2e test

john-a-joyce:
Initial config file changes to spawn
a CircleCI e2e test.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,85 @@ defaults: &defaults
     GOPATH: /go
     SKIP_CLEANUP: true
 
+# VM environment. Includes docker.
+integrationDefaults: &integrationDefaults
+  machine: true
+  working_directory: /go/src/istio.io/cni #same as docker
+  environment:
+    - CHANGE_MINIKUBE_NONE_USER: true
+    - GOPATH: /go
+    - SKIP_CLEANUP: true
+    - KUBECONFIG: /go/out/minikube.conf
+    - TEST_ENV: minikube-none
+
+# Common procedure to initialize working directory
+initWorkingDir: &initWorkingDir
+  type: shell
+  name: Initialize Working Directory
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -9,7 +9,87 @@ defaults: &defaults
...
+            - TEST_ENV: minikube-none
+            - GOPATH: /go
+    steps:
+      - <<: *initWorkingDir
+      - checkout
+      - attach_workspace:
+          at:  /go
+      - <<: *markJobStartsOnGCS
+      - run: make sync
+      - run: bin/testEnvRootMinikube.sh start
```
tiswanso: We can make these jobs pull istio/istio from this job definition and invoke the bin/testEnvRootMinikube.sh from there.  IMO, we shouldn't copy it into this repo.

john-a-joyce: Yeah - the job needs to pull istio/istio but my thinking we might have jobs that want to create a minikube environment without istio/istio being pulled.   I guess we can include that when it is needed.   For now I want to get some type of trigger in so it was easier to iterate on a couple different vectors.

.circleci/config.yml
```diff
@@ -9,7 +9,87 @@ defaults: &defaults
...
+      - <<: *markJobFinishesOnGCS
+      - store_artifacts:
+          path: /home/circleci/logs
+      - store_artifacts:
+          path: /tmp
+      - store_artifacts:
+          path: /var/lib/localkube/
+      - store_test_results:
+          path: /go/out/tests
+
```
tiswanso: need to reference this job from the workflows section to get it to run.  The "all" list runs on PRs.

tiswanso:
Closing this as John is on PTO and it conflicts with the install-cni circleci job PR #32 that merged.  All of this change was absorbed into PR #33.

PR #24, opened 2018-10-11
Add install-cni test 

tiswanso:
This is a combo implementation to establish a couple of test cases for the install-cni container.  It currently requires docker.  

Details:
* `make test` or `make install-test`
* docker runs install-cni with dirs mounted and env vars set
   * checks the resulting CNI plugin configuration and temp bin dir contents
   * does a straight file compare of plugin config
   * checks cleanup as well

PR #32, opened 2018-10-18
Create circleci job for install-cni test

tiswanso:
~depends on PR #24 .  Will rebase this once #24 merges.~  Runs test merged with PR #24 

Runs `make test` in a circleci job in a VM to be able to do the `docker run`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,6 +9,29 @@ defaults: &defaults
     GOPATH: /go
     SKIP_CLEANUP: true
 
+# VM environment. Includes docker.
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -22,11 +45,37 @@ jobs:
       - checkout
       - run: make lint
 
+  install-cni:
+    <<: *integrationDefaults
+    environment:
+            - GOPATH: /go
```
baodongli: the above two lines may be removed.

tiswanso: cool... done

PR #33, opened 2018-10-18
Add e2e-simple circleci

tiswanso:
update of PR #31 to pull istio and run e2e-simple (with istio TAG=nightly-master)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,7 +11,8 @@ defaults: &defaults
 
 # VM environment. Includes docker.
 integrationDefaults: &integrationDefaults
-  machine: true
+  machine:
+    image: circleci/classic:edge
   working_directory: /go/src/istio.io/cni #same as docker
   environment:
     - CHANGE_MINIKUBE_NONE_USER: true
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -159,7 +159,7 @@ jobs:
           working_directory: /go/src/istio.io/istio
           command: |
             HUB=gcr.io/istio-release TAG=master-latest-daily make istioctl
-            HUB=gcr.io/istio-release TAG=master-latest-daily ENABLE_ISTIO_CNI=true make test/local/auth/e2e_simple
+            HUB=gcr.io/istio-release TAG=master-latest-daily ENABLE_ISTIO_CNI=true E2E_ARGS="--kube_inject_configmap istio-sidecar-injector" make test/local/auth/e2e_simple
       - <<: *recordZeroExitCodeIfTestPassed
       - <<: *recordNonzeroExitCodeIfTestFailed
       - <<: *markJobFinishesOnGCS
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,6 +28,7 @@ initWorkingDir: &initWorkingDir
   command: |
     sudo mkdir -p /go/src/istio.io/cni
     sudo chown -R circleci /go
+    sudo mkdir -p /go/src/istio.io/istio
     mkdir -p /go/out/tests
     mkdir -p /go/out/logs
     mkdir -p /home/circleci/logs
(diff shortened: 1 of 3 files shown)


tiswanso:
Contains content from #32.  Will rebase this after #32 merges.  

The current e2e-simple failure looks like the single step `kubectl apply -f istio.yaml` problem (galley isn't ready to validate webhooks fast enough).  I _think_ the istio/istio circleci tests pass because they always build the docker images for istio so they come up very fast (no pull because images is already in docker).  Since this job is just grabbing istio nightly images and not building I think it needs to explicitly `docker pull` galley for now.  Will update and try.

tiswanso:
Pretty close to ready... need to fix the istio/istio deployment to allow for the following:
1.  bin/testEnvRootMinikube.sh -- add arg for CNI
1.  allow for setting istio-cni hub and tag via `make e2e_simple`
1.  pull in patch for making e2e_simple use istioctl kubeinject with configmap for sidecar inject to not inject initContainer

.circleci/config.yml
```diff
@@ -67,6 +95,64 @@ jobs:
...
+      - run:
+          command: |
+            if [ ! -f /go/out/linux_amd64/release/istio-cni ]; then
+              # Should only happen when re-running a job, and the workspace is gone
+              time make build
+            fi
+            make docker.all
+      - run: docker images
+      - run:
+          name: Get Istio
```
john-a-joyce: I had started a small sh script to do this.  It was in a different branch that i haven't submitted a PR for.  The reason I started this with a shell script is so that if someone wanted to run the tests locally it would be done in a uniform way.   Although re-considering since there is some complexity to doing it in a script and instead we can require the user checkout and build Istio anyway they want.

.circleci/config.yml
```diff
@@ -67,6 +95,64 @@ jobs:
...
+          shell: /bin/bash
+          working_directory: /go/src/istio.io/istio
+          command: |
+            git clone https://github.com/tiswanso/istio.git .
```
john-a-joyce: We should pick a specific Istio tag here (e.g. 1.0.2) and use istio/istio not a private repo.   I see below you don't build the Istio containers but pull from the master-latest-daily.  That is risking because when the test code and associated containers aren't in lock step failures are possible and probably over time.

.circleci/config.yml
```diff
@@ -67,6 +95,64 @@ jobs:
...
+      - run: /go/src/istio.io/istio/bin/testEnvRootMinikube.sh start
+      - run: /go/src/istio.io/istio/bin/testEnvRootMinikube.sh wait
+      - run: docker pull gcr.io/istio-release/galley:master-latest-daily
+      - run: docker images
+      - run:
+          name: Apply Istio CRDs
+          shell: /bin/bash
+          working_directory: /go/src/istio.io/istio
+          command: |
+            kubectl apply -f install/kubernetes/helm/istio/templates/crds.yaml
```
john-a-joyce: The tests do this step so it should be needed.

.circleci/config.yml
john-a-joyce: " bin/testEnvRootMinikube.sh -- add arg for CNI"   what arg are you thinking is needed?  When I went through the script I thought it could be used as is?   
   
    "pull in patch for making e2e_simple use istioctl kubeinject with configmap for sidecar inject to not inject initContainer"

That patch merged so latest master would have it. Also Steves patch is needed (which also merged).  Finally and most importantly your patch is needed which I don't think there is an open PR yet for.

tiswanso: RE: testEnvRootMinikube.sh -- minikube doesn't configure kubelet to use CNI by default.  It has a command line param that it doesn't look like that script sets when starting minikube.

RE: patches -- Right... that's why I used my private repo.

.circleci/config.yml
tiswanso: Agreed... but to do initial job setup and get things working I'm using a private repo to figure out all required stuff for istio/istio.

.circleci/config.yml
tiswanso: The tests don't do it as a separate step so there's a timing issue that I see all the time without this step.  I'm pretty sure the main reason is because the istio/istio circleci tests all build the docker images locally and don't need to pull but in istio-cni we just want to grab a known good istio/istio hub+tag and not build images locally.

tiswanso:
The status of this is that there's still something not quite right about the minikube setup for CNI added under

tiswanso:
Update--the minikube script starts minikube with `vm-driver=none` which doesn't automatically setup any CNI plugin when network-plugin=cni.  So, there is no primary CNI plugin setup by default.  This seems to cause all of the istio pods to not make it to running.  

Updated `github.com/tiswanso/istio.git` to configure the bridge CNI plugin.

.circleci/config.yml
john-a-joyce: Since the other changes merged should be able to pull master.  Even better is if there is a tag corresponding to the master-latest-daily images we are pulling.

.circleci/config.yml
```diff
@@ -86,6 +114,65 @@ jobs:
...
+      - run:
+          name: Run Istio e2e_simple
+          shell: /bin/bash
+          working_directory: /go/src/istio.io/istio
+          command: |
+            HUB=gcr.io/istio-release TAG=master-latest-daily make istioctl
+            HUB=gcr.io/istio-release TAG=master-latest-daily ENABLE_ISTIO_CNI=true EXTRA_HELM_SETTINGS="--set istio-cni.excludeNamespaces={} --set istio-cni.tag=${CIRCLE_SHA1}" E2E_ARGS="--kube_inject_configmap=istio-sidecar-injector" make test/local/auth/e2e_simple
```
john-a-joyce: Need to set istio-cni.hub=istio

tiswanso:
Closing this as the e2e strategy changed to use https://github.com/kubernetes-sigs/kubeadm-dind-cluster for test kubernetes clusters with a CNI plugin (flannel).  See PR #62

Issue #35, opened 2018-10-19
cmdAdd unable to get k8sArgs


tiswanso:
PR #26 modified the `K8sArgs` element names to fit go standard but that broke the reflect unmarshalling done by `types.LoadArgs()`

PR #36, opened 2018-10-19
fix k8sArgs & make circleci use go 1.10

tiswanso:
Fix k8sArgs elements to exactly match the kubelet key value pairs in the arg string so unmarshalling works.

- add a test case for checking unmarshalling with a sample kubelet arg string
- Add check for sidecar inject disable annotation to not do redirect (set by Istio control-plane pods)
- Changes to .circleci/config.yaml match istio/istio for setting up go 1.10 in the VM

Signed-off-by: Tim Swanson <tiswanso@cisco.com>

Closes: #35

### DIFF
--- a/cmd/istio-cni/main.go
+++ b/cmd/istio-cni/main.go
@@ -217,10 +217,8 @@ func cmdAdd(args *skel.CmdArgs) error {
 			if k8sErr != nil {
 				logger.Warnf("Error geting Pod data %v", k8sErr)
 			}
-			//for _, container := range containers {
-			//if container == "istio-proxy" {
 			logger.Infof("Found containers %v", containers)
-			if len(containers) >= 1 {
+			if len(containers) > 1 {
 				logrus.WithFields(logrus.Fields{
 					"ContainerID": args.ContainerID,
 					"netns":       args.Netns,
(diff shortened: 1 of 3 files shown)


cmd/istio-cni/main_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/containernetworking/cni/pkg/types"
+)
+
+func TestLoadArgs(t *testing.T) {
+	kubeletArgs := "IgnoreUnknown=1;K8S_POD_NAMESPACE=istio-system;" +
+		"K8S_POD_NAME=istio-sidecar-injector-8489cf78fb-48pvg;" +
+		"K8S_POD_INFRA_CONTAINER_ID=3c41e946cf17a32760ff86940a73b06982f1815e9083cf2f4bfccb9b7605f326"
+
+	k8sArgs := K8sArgs{}
```
rcurran1: consider adding a comment here that this Test is testing the K8sArgs d.s. Here or at top of function.

cmd/istio-cni/main.go
```diff
@@ -221,12 +224,24 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+					"pod":         string(k8sArgs.K8S_POD_NAME),
+					"Namespace":   string(k8sArgs.K8S_POD_NAMESPACE),
 					"ports":       ports,
+					"annotations": annotations,
 				}).Infof("Updating iptables redirect for Istio proxy")
-
-				_ = setupRedirect(args.Netns, ports)
+				if val, ok := annotations[injectAnnotationKey]; ok {
+					logrus.Infof("Pod %s contains inject annotation: %s", string(k8sArgs.K8S_POD_NAME), val)
+					if b, err := strconv.ParseBool(val); err == nil {
```
rcurran1: consider change 'b' to something more meaningful.

cmd/istio-cni/main.go
```diff
@@ -221,12 +224,24 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+						if !b {
+							logrus.Infof("Pod excluded due to annotations")
+							excludePod = true
+						}
+					}
```
rcurran1: do we need logs here for 'not ok' and 2nd conditional failed 'err'?
i.e.
} else {
       logrus.Infof ("parsebool failed for value %s err: %v", val, err)

cmd/istio-cni/main.go
tiswanso: done

Issue #42, opened 2018-10-24
Add exclude/include IP ranges capability


tiswanso:
Match the pattern of the istio sidecar injector for params to the istio-iptables.sh script.

The e2e_simple test's nginx test requires support for the annotations mapped to these params.

tiswanso:
The Istio CNI plugin needs to map the pod annotations to the istio-iptables.sh vars like the sidecar-injector template does:  
https://github.com/istio/istio/blob/master/pilot/pkg/kube/inject/mesh.go#L47-L56

PR #43, opened 2018-10-25
Map pod annotation data to redirect params

tiswanso:
If annotation keys `traffic.sidecar.*` are present on the pod, use their values as params. 

~TODO--validate values in annotations~ Done

Closes: #42

### DIFF
--- a/cmd/istio-cni/redirect.go
+++ b/cmd/istio-cni/redirect.go
@@ -1,0 +1,268 @@
+// Copyright 2018 Istio authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Defines the redirect object and operations.
+package main
+
+import (
+	"fmt"
+	"net"
+	"os/exec"
+	"strconv"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	redirectModeREDIRECT         = "REDIRECT"
+	redirectModeTPROXY           = "TPROXY"
+	defaultProxyStatusPort       = "15020"
+	defaultRedirectToPort        = "15001"
+	defaultNoRedirectUID         = "1337"
+	defaultRedirectMode          = redirectModeREDIRECT
+	defaultRedirectIPCidr        = "*"
+	defaultRedirectExcludeIPCidr = ""
+	defaultRedirectExcludePort   = defaultProxyStatusPort
+
+	includeIPCidrsKey = "traffic.sidecar.istio.io/includeOutboundIPRanges"
+	excludeIPCidrsKey = "traffic.sidecar.istio.io/excludeOutboundIPRanges"
+	includePortsKey   = "traffic.sidecar.istio.io/includeInboundPorts"
+	excludePortsKey   = "traffic.sidecar.istio.io/excludeInboundPorts"
+
+	sidecarInterceptModeKey = "sidecar.istio.io/interceptionMode"
+	sidecarPortListKey      = "status.sidecar.istio.io/port"
+	sidecarStatusKey        = "sidecar.istio.io/status"
+)
+
+var (
+	annotationRegistry = map[string]*annotationParam{
+		"inject":         {injectAnnotationKey, "", alwaysValidFunc},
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -72,6 +100,64 @@ jobs:
...
+              time make build
+            fi
+            make docker.all
+      - run: docker images
+      - run:
+          name: Get Istio
+          shell: /bin/bash
+          working_directory: /go/src/istio.io/istio
+          command: |
+            git clone https://github.com/tiswanso/istio.git .
```
john-a-joyce: Since the Circle CI is not going to pass anyway you might want to change this istio/istio and checkout master below

.circleci/config.yml
```diff
@@ -72,6 +100,64 @@ jobs:
...
+          working_directory: /go/src/istio.io/istio
+          command: |
+            kubectl apply -f install/kubernetes/helm/istio/templates/crds.yaml
+      - run:
+          name: Run Istio e2e_simple
+          shell: /bin/bash
+          working_directory: /go/src/istio.io/istio
+          command: |
+            HUB=gcr.io/istio-release TAG=master-latest-daily make istioctl
+            HUB=gcr.io/istio-release TAG=master-latest-daily ENABLE_ISTIO_CNI=true EXTRA_HELM_SETTINGS="--set istio-cni.excludeNamespaces={} --set istio-cni.tag=v0.1-dev" E2E_ARGS="--kube_inject_configmap=istio-sidecar-injector" make test/local/auth/e2e_simple
```
john-a-joyce: Interesting paradox here.  Should change the v0.10dev label to match the label corresponding to this PR.   Again since the test doesn't pass maybe it's not critical

john-a-joyce:
Also I would like to see these changes merge sooner than later because it moves the test code much further along in case there is any parallel work that can be done.

.circleci/config.yml
baodongli: Just a thought. Instead of duplicating a lot of stuff for e2e or integration tests, maybe these tests can be incorporated into the istio repo. This repo will simply perform unit tests. 

In that case, istio.io/cni will be brought into istio.io/istio as a vendor dependency.

john-a-joyce: @baodongli - we aren't duplicating anything we are just using istio/istio e2e directly with a few tweaks.   Most of the tweaks would be required even without e2e since they are needed to have a CNI deploy option.   I think the approach above is cleaner than having a vendor dependency because it constrains the dependency to only the gate testing.   The difference might be subtle though so either approach could be workable.

tiswanso: This comment thread really should be on #33 ... this PR is just rebased on top of it

baodongli: Yeah, I realized that. A few PRs contain similar change.

baodongli: Wonder if you can just modify your PR so that only changes pertaining to this PR is included. Barring changes in the same file from other PRs, When you commit, you can just commit your change only, and push them up. It seems to be possible in your case.

tiswanso: Yeah, I usually do that but have it on top of this to get e2e_simple to pass.  But circleci isn't hooked up properly now after the repo move and, as @john-a-joyce points out, the e2e test isn't using the right tag anyway for each specific istio/cni change.  I'll go ahead and rebase it on master to get rid of the other diffs.

.circleci/config.yml
tiswanso: the test should pass with these changes.  It passes locally.  But yeah, we need to move the tag to the build that's happening for each PRs diffs.

cmd/istio-cni/redirect.go
```diff
@@ -0,0 +1,268 @@
...
+var (
+	annotationRegistry = map[string]*annotationParam{
+		"inject":         {injectAnnotationKey, "", alwaysValidFunc},
+		"status":         {sidecarStatusKey, "", alwaysValidFunc},
+		"redirectMode":   {sidecarInterceptModeKey, defaultRedirectMode, validateInterceptionMode},
+		"ports":          {sidecarPortListKey, "", validatePortList},
+		"includeIPCidrs": {includeIPCidrsKey, defaultRedirectIPCidr, validateCIDRListWithWildcard},
+		"excludeIPCidrs": {excludeIPCidrsKey, defaultRedirectExcludeIPCidr, validateCIDRList},
+		"includePorts":   {includePortsKey, "", validatePortListWithWildcard},
+		"excludePorts":   {excludePortsKey, defaultRedirectExcludePort, validatePortList},
```
baodongli: Wonder why using short key such as "inject", "status", etc.
Why not just define the map as
 annotationRegistry = map[string]*annotationParams{
injectAnnotationKey: {"", alwaysValidFunc},
......
}

If it's necessary to use the short key, it seems to be appropriate to define them as constants as well.

tiswanso:
fyi, this passes e2e_simple on baremetal and also was used for e2e circleci DinD testing:  https://circleci.com/gh/istio/cni/755

PR #53, opened 2018-10-31
Adding nightly

costinm:
Once this completes, I'll update the image references to point to istionightly.

I think we can combine lint/build in a single circleci job - in istio/istio things are super slow and need to be in parallel, hopefully we'll not end up in the same state in this repo.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,7 +80,6 @@ jobs:
       - attach_workspace:
           at:  /go
       - run: GOOS=linux make build
-      - run: make lint
       - run: make test
       - run:
           command: |
@@ -112,5 +111,6 @@ workflows:
             only:
             - master
     jobs:
+    - lint
     - nightly:
         context: org-global
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -106,7 +106,7 @@ workflows:
   nightly:
     triggers:
     - schedule:
-        cron: "30 23 * * *"
+        cron: "30 * * * *"
         filters:
           branches:
             only:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -114,5 +114,3 @@ workflows:
     jobs:
     - nightly:
         context: org-global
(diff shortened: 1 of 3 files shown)


tiswanso:
Can someone with the merge button merge this?

Not sure why helm lint is failing like this:

.circleci/config.yml
```diff
@@ -72,6 +72,25 @@ jobs:
       - store_test_results:
           path: /go/out/tests
 
+  nightly:
+    <<: *integrationDefaults
```
tiswanso: looks like this needs to be `*defaults` to use the docker image with helm so `make lint` will work.  We should probably just make this `nightly` job the docker push job and add the `lint` and `install-cni` jobs to the "nightly" workflow section.

.circleci/config.yml
```diff
@@ -82,3 +101,16 @@ workflows:
...
+    triggers:
+    - schedule:
+        cron: "30 * * * *"
+        filters:
+          branches:
+            only:
+            - master
+    jobs:
+    - nightly:
+        context: org-global
```
tiswanso: I think we need to make this like istio/istio has it with the "nightly" job just doing the stuff required for docker push but the "nightly" workflow invoking the other jobs as well--lint, install-cni (test), etc.

```
    jobs:
      - lint
      - build
      - install-cni:
          requires:
            - build
        - nightly:
             context: org-global
             requires:
              - build
```

.circleci/config.yml
costinm: Move lint to separate job in the workflow.

I'll iterate - there are few other optimizations we can make, first step is to get things moving.

.circleci/config.yml
costinm: Yes, or the other way around. Lint/build are separate in circle because they're very slow.

.circleci/config.yml
tiswanso: sounds good

PR #61, opened 2018-11-05
Fix unit test failure in main_test.go

rcurran1:


cmd/istio-cni/main.go
```diff
@@ -45,8 +45,8 @@ var (
 	injectAnnotationKey   = "sidecar.istio.io/inject"
 )
 
-// execCommand is a unit test override variable.
-var execCommand = exec.Command
+// setupRedirect is a unit test override variable.
```
venilnoronha: nit: prefer a more descriptive comment explaining this variable's behavior.

PR #62, opened 2018-11-07
CircleCI DinD job

tiswanso:
CircleCI with k8s on DinD similar to KDC repo--https://github.com/kubernetes-sigs/kubeadm-dind-cluster

1. use circleci remote_docker in the same pattern as KDC's circleci jobs to install a multi-node k8s cluster with flannel CNI.
1. save/restore install-cni docker image under-test from build job.  Copy images into each k8s node's docker cache
1. build a container with Istio master to run Istio e2e tests from the same docker network space as the DinD k8s' kube-master container ("node").  see run_e2e.sh
    1. run e2e_simple tests with Istio CNI enabled
1.  collect and copy out k8s state to circleci artifacts

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,6 +11,7 @@ defaults: &defaults
 
 dind-env: &dind-env
   DOCKER_VERSION: "17.03.0-ce"
+  CNI_PLUGIN: flannel
 
 # dind docker env
 dind-defaults: &dind-defaults
@@ -174,7 +175,7 @@ jobs:
       - run:
           name: Run e2e test
           command: |
-            docker run --name=istio-e2e --network=container:kube-master -t -e CNI_TAG='v0.1-dev-mapann_rebase' -e CNI_HUB='docker.io/tiswanso' -e HUB='docker.io/tiswanso' -e TAG='nov12' -e SKIP_CLEAN='1' e2e-runner:latest /go/run_e2e.sh
+            docker run --name=istio-e2e --network=container:kube-master -t -e CNI_TAG='v0.1-dev-mapann_rebase' -e CNI_HUB='docker.io/tiswanso' -e HUB='gcr.io/istio-release' -e TAG='master-latest-daily' -e SKIP_CLEAN='1' e2e-runner:latest /go/run_e2e.sh
       - run:
           name: Display cluster status (2)
           when: always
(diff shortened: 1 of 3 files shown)


tiswanso:
IMO, this should be ready to merge as an e2e test to validate istio/cni changes.

.circleci/config.yml
```diff
@@ -9,6 +9,63 @@ defaults: &defaults
...
+    SKIP_CLEANUP: true
+    <<: *dind-env
+  working_directory: ~/kubeadm-dind-cluster
+
+dind-setup: &dind-setup
+  name: Set up the environment
+  command: |
+    sudo apt-get -qq update
+    sudo apt-get install -y curl ca-certificates git liblz4-tool rsync socat tzdata
+    #curl -sSL -o "/tmp/docker-${DOCKER_VERSION}.tgz" "https://get.docker.com/builds/Linux/x86_64/docker-${DOCKER_VERSION}.tgz"
```
john-a-joyce: These look like leftover debugging comments.

.circleci/run_e2e.sh
```diff
@@ -0,0 +1,13 @@
...
+CNI_TAG=${CNI_TAG:-$1}
+
+cd /go/src/istio.io/istio
+
+HUB=${HUB:-gcr.io/istio-release}
+TAG=${TAG:-master-latest-daily}
+
+HUB=${HUB} TAG=${TAG} make istioctl
+
+HUB=${HUB} TAG=${TAG} ENABLE_ISTIO_CNI=true EXTRA_HELM_SETTINGS="--set istio-cni.excludeNamespaces={} --set istio-cni.hub=${CNI_HUB} --set istio-cni.tag=${CNI_TAG} --set istio-cni.pullPolicy=IfNotPresent --set istio-cni.logLevel=${CNI_LOGLVL:-debug}" E2E_ARGS="--kube_inject_configmap=istio-sidecar-injector ${SKIP_CLEAN:+ --skip_cleanup}" make test/local/auth/e2e_simple
```
john-a-joyce: --set istio-cni.pullPolicy=IfNotPresent - do we need override in this environment?

tiswanso: Yeah it needs to be overridden because otherwise it actually tries to pull the image from docker.io and it's not uploaded there.  It needs to use the image in the nodes' docker image cache that gets loaded with the `docker load -i ...`

.circleci/config.yml
tiswanso: ok will remove

PR #65, opened 2018-11-14
Run nightly only at 09:45am UTC and include e2e test

tiswanso:
change it from running every hour on the half past mark.  Only run it nightly at 09:45:00 UTC to try to ensure it picks up Istio's nightly release for testing--istio nightly seems to run at ~09:15 UTC.

Also include the DinD e2e test in the nightly run.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -133,6 +133,10 @@ jobs:
       - store_test_results:
           path: /go/out/tests
 
+  # nightly test & publish
+  # NOTE: publish only happens for CircleCI orgs with DOCKER_USER & DOCKER_PASS set.
+  #       Istio currently does not use CircleCI to publish nightly.  Nightly artifacts
+  #       are published by istio/istio/release tooling & go to gcr.io/istio-release.
   nightly:
     <<: *integrationDefaults
     steps:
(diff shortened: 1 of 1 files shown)


john-a-joyce:
Since we are cleaning up we should remove the "docker push" from the nightly job since it is not doing anything anyway or at least comment it out for now.

tiswanso:
>Since we are cleaning up we should remove the "docker push" from the nightly job since it is not doing anything anyway or at least comment it out for now.

@john-a-joyce I was thinking of that but I thought of how Costin setup his fork to send webhook triggers to his own circleci proj.  With the PoCs folks are intending to get going maybe they want to do something similar, setup those circleci org vars, and have their teams pull nightly images from their own nightly repos.  I don't see any harm in keeping it in there other than folks thinking there's a nightly image in docker.io/istio.  In any event, I'm intending to create a TESTING.md to explain the tests in this repo and in istio/istio and we can clarify this info on the nightly job.

john-a-joyce:
It's not a big deal if you want to leave it but please add comments explaining what really happens.   Regarding the TESTING.MD there is already a tests/README.md that started to cover that.  It is stale now and would need an update.

PR #71, opened 2018-11-20
Fix build info and other references to user tiswanso in misc places.

tiswanso:
leaving the references in the helm defaults (values.yaml) since we don't yet have a process that pushes to gcr.io/istio-release yet.


README.md
```diff
@@ -142,7 +142,7 @@ $ ISTIO_CNI_RELPATH=github.com/some/cni GOOS=linux make build
 To push the Docker image:
 
 ```console
-$ export HUB=docker.io/tiswanso
+$ export HUB=docker.io/myuser
```
venilnoronha: `$ export HUB=docker.io/$USER`

PR #75, opened 2019-01-21
Check for Sidecar Annotation Before Redirect

joshw123:
Hello, 

We had a scenario in our cluster were a pod got IP Rules when it did not have the sidecar, to prevent this I have added a check to make sure that the pod has a status annotation, which will only be present if there is a sidecar.

Feedback Welcome! 

Thank You


joshw123:
I signed it!

joshw123:
https://github.com/istio/cni/pull/75

joshw123:
I signed it!

PR #76, opened 2019-01-21
Check for Sidecar Annotation Before Redirect

joshw123:
Hello,
We had a scenario in our cluster were a pod got IP Rules when it did not have the sidecar, to prevent this I have added a check to make sure that the pod has a status annotation, which will only be present if there is a sidecar.
Feedback Welcome!
Thank You

### DIFF
--- a/cmd/istio-cni/main.go
+++ b/cmd/istio-cni/main.go
@@ -37,7 +37,7 @@ var (
 	nsSetupProg   = "istio-iptables.sh"
 
 	injectAnnotationKey = "sidecar.istio.io/inject"
-	injectIstioKey      = "sidecar.istio.io/status"
+	sidecarStatusKey    = "sidecar.istio.io/status"
 )
 
 // setupRedirect is a unit test override variable.
@@ -204,8 +204,8 @@ func cmdAdd(args *skel.CmdArgs) error {
 						}
 					}
 				}
-				if val, ok := annotations[injectIstioKey]; !ok {
-					logrus.Infof("Pod %s excluded due to not containing sidecar annotation: %s", string(k8sArgs.K8S_POD_NAME), val)
+				if _, ok := annotations[sidecarStatusKey]; !ok {
+					logrus.Infof("Pod %s excluded due to not containing sidecar annotation", string(k8sArgs.K8S_POD_NAME))
 					excludePod = true
 				}
 				if !excludePod {
--- a/cmd/istio-cni/main.go
+++ b/cmd/istio-cni/main.go
@@ -37,6 +37,7 @@ var (
 	nsSetupProg   = "istio-iptables.sh"
 
 	injectAnnotationKey = "sidecar.istio.io/inject"
+	injectIstioKey      = "sidecar.istio.io/status"
 )
 
 // setupRedirect is a unit test override variable.
@@ -202,6 +203,10 @@ func cmdAdd(args *skel.CmdArgs) error {
 							excludePod = true
 						}
 					}
+				}
+				if val, ok := annotations[injectIstioKey]; !ok {
+					logrus.Infof("Pod %s excluded due to not containing sidecar annotation: %s", string(k8sArgs.K8S_POD_NAME), val)
+					excludePod = true
 				}
 				if !excludePod {
 					logrus.Infof("setting up redirect")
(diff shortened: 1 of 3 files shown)


cmd/istio-cni/main.go
```diff
@@ -37,6 +37,7 @@ var (
 	nsSetupProg   = "istio-iptables.sh"
 
 	injectAnnotationKey = "sidecar.istio.io/inject"
+	injectIstioKey      = "sidecar.istio.io/status"
```
jwendell: Maybe move `sidecarStatusKey` declaration in redirect.go to here and use it instead?

cmd/istio-cni/main.go
```diff
@@ -203,6 +204,10 @@ func cmdAdd(args *skel.CmdArgs) error {
 						}
 					}
 				}
+				if val, ok := annotations[injectIstioKey]; !ok {
+					logrus.Infof("Pod %s excluded due to not containing sidecar annotation: %s", string(k8sArgs.K8S_POD_NAME), val)
```
jwendell: val here is empty. Maybe use `annotations[injectIstioKey]` instead?

cmd/istio-cni/redirect.go
```diff
@@ -48,14 +48,15 @@ const (
...
-		"status":         {sidecarStatusKey, "", alwaysValidFunc},
-		"redirectMode":   {sidecarInterceptModeKey, defaultRedirectMode, validateInterceptionMode},
-		"ports":          {sidecarPortListKey, "", validatePortList},
-		"includeIPCidrs": {includeIPCidrsKey, defaultRedirectIPCidr, validateCIDRListWithWildcard},
-		"excludeIPCidrs": {excludeIPCidrsKey, defaultRedirectExcludeIPCidr, validateCIDRList},
-		"includePorts":   {includePortsKey, "", validatePortListWithWildcard},
-		"excludePorts":   {excludePortsKey, defaultRedirectExcludePort, validatePortList},
+		"inject":          {injectAnnotationKey, "", alwaysValidFunc},
+		"injectNoSidecar": {injectIstioKey, "", alwaysValidFunc},
+		"status":          {sidecarStatusKey, "", alwaysValidFunc},
```
jwendell: Duplicate of sidecarStatusKey?

joshw123:
@jwendell Thank You for the feedback! I have committed the changes

cmd/istio-cni/main_test.go
```diff
@@ -207,6 +207,7 @@ func TestCmdAddTwoContainers(t *testing.T) {
 
 	setupRedirect = mockNsenterRedirect
 	testAnnotations[injectAnnotationKey] = "true"
+	testAnnotations[sidecarStatusKey] = "true"
```
venilnoronha: Should `testAnnotations[sidecarStatusKey]` be set to `true` by default and only set to `false` in the `TestCmdAddTwoContainersWithoutSideCar` test?

joshw123: Hi @venilnoronha I'm not sure that this will work, The value of the annotation will be set to false, but the annotation is still set so it would still try and redirect as the annotation is present, even without the sidecar.

venilnoronha: My concern here is that all tests might be affected by the `excludePod` logic based on the status annotation. Can we still set the annotation by default and remove it just for a specific test?

cmd/istio-cni/main_test.go
```diff
@@ -216,6 +217,18 @@ func TestCmdAddTwoContainers(t *testing.T) {
 	}
 }
 
+func TestCmdAddTwoContainersWithoutSideCar(t *testing.T) {
+	defer resetGlobalTestVariables()
+
+	testAnnotations = nil
```
venilnoronha: I think it'd be better to do `testAnnotations = map[string]string{}` just like in the `resetGlobalTestVariables` function.

venilnoronha: Actually, it'd be much better to do `delete(testAnnotations, sidecarStatusKey)` instead.

joshw123: That makes a lot more sense! Thank You. I have commited the change 👍

venilnoronha:
Can you amend your commit/add a dummy commit to kick CI?

venilnoronha:
@joshw123 thanks for the contribution!

joshw123:
Thank You @venilnoronha

joshw123:
@venilnoronha Do you know when this will be released? Thank You!

venilnoronha:
I'm hoping Istio 1.1.

venilnoronha:
I've updated the CNI package in istio/istio#11249.

joshw123:
Thank You

Issue #82, opened 2019-02-02
Race-condition when starting istio-cni before calico or canal


jeckhart:
Our dev environment is a kops (v1.11.0) deployed kubernetes cluster (v1.11.7) with canal (calico + flannel).

We are using istio-cni (0.1-dev). When starting a new node istio-cni reliably starts faster than calico-cni.

This results in istio-cni creating a config file (in this case `10-canal.conflist`) that is later overwritten by the calico-cni startup. The end result is that istio-cni is not deployed because it's config is overwritten.

I've updated the startup command for istio-cni to wait for the conf file to exist before starting. The relevant parts look like:
```
apiVersion: extensions/v1beta1
kind: DaemonSet
spec:
  template:
    spec:
      containers:
      - args:
        - -c
        - while [[ -n "${CNI_CONF_NAME}" ]] && [[ ! -e "/host/etc/cni/net.d/${CNI_CONF_NAME}"
          ]]; do echo Waiting for cni to stabilize; sleep 1; done; /install-cni.sh
        command:
        - /bin/bash
...
```

I'm looking for feedback on how others are handling this. Is anybody else encountered a race with istio-cni? Are there other solutions? If not, should I submit this as a PR (adding an option to make if configurable)?

dicarlo2:
https://github.com/istio/istio/issues/11037

tiswanso:
@dicarlo2 -- thanks for troubleshooting and filing this issue...  Good catch!  

We could enhance the install-cni.sh [1] to periodically ensure that the istio-cni config is present AND at the end of the plugin config list to fix this.  

A work-around would be to restart the istio-cni pod on the node (kill the pod so a k8s starts a new istio-cni pod).  That would cause the install-cni.sh script to be rerun on the node and reapply the istio-cni CNI plugin config to the calico CNI config file.

[1] https://github.com/istio/cni/blob/master/deployments/kubernetes/install/scripts/install-cni.sh

tiswanso:
@dicarlo2 -- I pushed PR #83 with a fix to detect a few conditions of istio-cni CNI config getting clobbered or superseded and causing a pod restart to reapply the config.  This should resolve your issue.  The general istio-cni lint check is failing at the moment but you can try the fix out via using the istio-cni image:
```
hub=docker.io/tiswanso
tag=install-race  
```

OR use the following helm chart settings with the istio-cni chart: 
```
istio-cni.hub=docker.io/tiswanso
istio-cni.tag=install-race
```

[...]

PR #83, opened 2019-02-04
Detect clobbering of istio-cni content in CNI conf file

tiswanso:
Modify istio-cni-node daemonset's install-cni.sh to periodically check for
- removed istio-cni content from plugins list
- removal of CNI config file
- adding of CNI config file that supersedes the on with istio-cni content
  (lexicographically first file)

If found, exit the proc with error so k8s will restart the pod and the
istio-cni install logic will reapply the config.

Fixes: #82

### DIFF
--- a/deployments/kubernetes/install/scripts/install-cni.sh
+++ b/deployments/kubernetes/install/scripts/install-cni.sh
@@ -28,17 +28,26 @@ function rm_bin_files() {
 HOST_CNI_NET_DIR=${CNI_NET_DIR:-/etc/cni/net.d}
 MOUNTED_CNI_NET_DIR=${MOUNTED_CNI_NET_DIR:-/host/etc/cni/net.d}
 
+CNI_CONF_NAME_OVERRIDE=${CNI_CONF_NAME:-}
+
 # default to first file in `ls` output
+# if dir is empty, default to a filename that is not likely to be lexicographically first in the dir
 CNI_CONF_NAME=${CNI_CONF_NAME:-$(ls ${MOUNTED_CNI_NET_DIR} | head -n 1)}
-CNI_CONF_NAME=${CNI_CONF_NAME:-10-calico.conflist}
+CNI_CONF_NAME=${CNI_CONF_NAME:-YYY-istio-cni.conflist}
 KUBECFG_FILE_NAME=${KUBECFG_FILE_NAME:-ZZZ-istio-cni-kubeconfig}
 CFGCHECK_INTERVAL=${CFGCHECK_INTERVAL:-1}
 
 function check_install() {
   cfgfile_nm=$(ls ${MOUNTED_CNI_NET_DIR} | head -n 1)
   if [[ "${cfgfile_nm}" != "${CNI_CONF_NAME}" ]]; then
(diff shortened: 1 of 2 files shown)


deployments/kubernetes/install/scripts/install-cni.sh
```diff
@@ -32,6 +32,25 @@ MOUNTED_CNI_NET_DIR=${MOUNTED_CNI_NET_DIR:-/host/etc/cni/net.d}
...
+  fi
+  if [ -e "${MOUNTED_CNI_NET_DIR}/${CNI_CONF_NAME}" ]; then
+    istiocni_conf=$(cat ${MOUNTED_CNI_NET_DIR}/${CNI_CONF_NAME} | jq '.plugins[]? | select(.type == "istio-cni")')
+    if [[ "$istiocni_conf" == "" ]]; then
+      echo "ERROR: istio-cni CNI config removed from file: \"${MOUNTED_CNI_NET_DIR}/${CNI_CONF_NAME}\""
+      exit 1  
+    fi
+  else
+    echo "ERROR: CNI config file \"${MOUNTED_CNI_NET_DIR}/${CNI_CONF_NAME}\" removed."
+    exit 1
```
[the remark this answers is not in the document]
tiswanso: Hi Venil,   It will loop forever if nothing changes just like it does today.  But if any of these checks in `check_install` hit then it will exit the proc and k8s will restart the pod.  When the pod restarts, the install-cni.sh config apply steps of the script will rerun and the istio-cni config will be applied again and go back into the check_install() loop.

So, you will be able to tell when something messed up the istio-cni config by how many times the `istio-cni-node` pods restart on a node.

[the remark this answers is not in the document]
venilnoronha: Thanks for the clarification.

deployments/kubernetes/install/scripts/install-cni.sh
```diff
@@ -32,6 +32,25 @@ MOUNTED_CNI_NET_DIR=${MOUNTED_CNI_NET_DIR:-/host/etc/cni/net.d}
 CNI_CONF_NAME=${CNI_CONF_NAME:-$(ls ${MOUNTED_CNI_NET_DIR} | head -n 1)}
 CNI_CONF_NAME=${CNI_CONF_NAME:-10-calico.conflist}
 KUBECFG_FILE_NAME=${KUBECFG_FILE_NAME:-ZZZ-istio-cni-kubeconfig}
+CFGCHECK_INTERVAL=${CFGCHECK_INTERVAL:-1}
```
venilnoronha: I think a default of `10` would be better for `sleep`, just like before.

tiswanso: I am thinking we need a smaller default to shorten the window of misconfig for the cases highlighted by issue #82 and Spike's comment here:  https://github.com/istio/istio/issues/11037#issuecomment-460368114

The gist is that we can't ensure the init order of the pods that do the per-node CNI installations (like istio-cni-node) when adding new nodes to the cluster AND we can't ensure control of when user pods start getting scheduled to the newly added nodes.

venilnoronha: Got it.

PR #86, opened 2019-02-15
Look for .conflist file if .conf doesn't exist

### LINKED ISSUE #85: Redeploying the istio-cni-installer pod breaks the net.d .conf file
If the pod that installs istio-cni gets redeployed for any reason, and the original net.d file ends with .conf, the end result is a broken .conflist file that no longer contains the original CNI plugin:
```
{
  "plugins": [
    {
      "type": "istio-cni",
      "log_level": "info",
      "kubernetes": {
        "kubeconfig": "/etc/cni/net.d/ZZZ-istio-cni-kubeconfig",
        "cni_bin_dir": "/opt/cni/bin",
        "exclude_namespaces": [
          "istio-system"
        ]
      }
    }
  ],
  "name": "k8s-pod-network",
  "cniVersion": "0.3.0"
}
```

This happens because the install script renames the file to .conflist. When the pod is redeployed, it can no longer find the original .conf file (which `CNI_CONF_FILE` points to), so it simply creates a new file with only the istio-cni plugin. The original main CNI plugin is not listed anywhere anymore.

luksa:
Fixes #85

### DIFF
--- a/deployments/kubernetes/install/scripts/install-cni.sh
+++ b/deployments/kubernetes/install/scripts/install-cni.sh
@@ -202,6 +202,13 @@ echo "CNI config: $(cat ${TMP_CONF})"
 sed -i s/__SERVICEACCOUNT_TOKEN__/${SERVICEACCOUNT_TOKEN:-}/g $TMP_CONF
 
 CNI_CONF_FILE=${MOUNTED_CNI_NET_DIR}/${CNI_CONF_NAME}
+
+if [ ! -e "${MOUNTED_CNI_NET_DIR}/${CNI_CONF_NAME}" ] && [ "${CNI_CONF_NAME: -5}" == ".conf" ] && [ -e "${MOUNTED_CNI_NET_DIR}/${CNI_CONF_NAME}list" ]; then
+    echo "${CNI_CONF_NAME} doesn't exist, but ${CNI_CONF_NAME}list does; Using it instead."
+    CNI_CONF_NAME="${CNI_CONF_NAME}list"
+	CNI_CONF_FILE=${MOUNTED_CNI_NET_DIR}/${CNI_CONF_NAME}
+fi
+
 if [ -e "${MOUNTED_CNI_NET_DIR}/${CNI_CONF_NAME}" ]; then
     # This section overwrites an existing plugins list entry to for istio-cni
     CNI_TMP_CONF_DATA=$(cat ${TMP_CONF})


luksa:
You might argue that istio-cni cleanup should revert the file to its original name & contents, but in reality, the file might also get updated (and renamed) by another agent (not related to istio at all). 

Basically, if the other agent renames the file from `.conf` to `.conflist` before the istio-cni installer runs (and assuming `CNI_CONF_FILE` is set to `something.conf`), the istio-cni installer needs to look for .conflist even though `CNI_CONF_FILE` points to `.conf`

tiswanso:
I'm wondering if we should even be renaming the file to ".conflist".  Maybe this fix should just be to remove that.  

Also, IMO, I don't think starting iinstall-cni.sh with env `CNI_CONF_NAME` (helm chart `cniConfFileName`) is advisable.  The behavior of kubelet is that the CNI conf file is found by using the lexicographical first file in the CNI config dir.  The install-cni.sh will do the same thing.  

In #83, I make the daemonset detect clobbering of the CNI conf and in that check it detects whether the file with the istio-cni config is the first file in the CNI conf dir.  If the env `CNI_CONF_NAME` is set for install-cni.sh it will just echo a warning, but if env `CNI_CONF_NAME` is not set it will restart the proc and the config will be applied to the new lexicographical first file.  This case will work if the install-cni.sh creates the CNI conf file prior to another CNI plugin's installer since the file istio-cni's install-cni.sh defaults to a file not likely to be lexicographical first.

luksa:
> I'm wondering if we should even be renaming the file to ".conflist". Maybe this fix should just be to remove that.

I've just checked this. It must be a .conflist file. See https://github.com/kubernetes/kubernetes/blob/954996e231074dc7429f7be1256a579bedd8344c/pkg/kubelet/dockershim/network/cni/cni.go#L158-L163


> Also, IMO, I don't think starting iinstall-cni.sh with env `CNI_CONF_NAME` (helm chart `cniConfFileName`) is advisable. The behavior of kubelet is that the CNI conf file is found by using the lexicographical first file in the CNI config dir. The install-cni.sh will do the same thing to detect the conf file if env `CNI_CONF_NAME` is unset.

I agree, but the Kubelet doesn't just use the lexicographical first file. It uses the first _valid_ file (it logs a warning if the first file isn't ok and then moves on to the next file).  Install-cni.sh should thus do the same.

tiswanso:
@luksa Great catch!   Yeah, I agree we should follow those semantics.

So the logic is:
```
foreach file in cni_conf_dir:
   if file suffix is ".conflist":
      if it contains a valid conflist:
           use it
      else:
          continue to next file
   else:
      if it contains a top level key "type" with a value:
          use it (non conflist case)
      else:
          continue to next file
```

For now, we could implement that basic logic within the bash script using `cat <file> | jq <check contents>` but, ideally, we'd convert the install-cni.sh to golang to use those libcni methods to absorb the file.  Still there's a decent bit of logic up in kubelet so even the golang approach would have a good amount of duplicate logic.

Would you be able to work on implementing the logic in install-cni.sh for this in this PR?

tiswanso:
I rebased this locally and basic testing seems fine.  If you could rebase this so CI passes we can get this in.  I'll do the other logic we discussed in a separate PR

luksa:
@tiswanso Rebased to latest master.

PR #96, opened 2019-03-08
install-cni: properly detect CNI conf file (match kubelet approach)

tiswanso:
Adds the same logic as kubelet to find the CNI config file in the CNI conf dir, rather than just look for the lexicographical first file. 
- Add test cases for skipping invalid .conf and .conflist files
- Add test case for CNI_CONF_FILE env var pointing to a missing .conf file but the already converted .conflist file is present. (issue #86)


tiswanso:
@luksa  @john-a-joyce -- Here's the fix for adding the logic to properly detect the CNI config file (basically match what kubelet does).  

Also, I added a bunch more test cases for the install-cni.sh logic.  (of course that took me about 3 times longer than the actual install-cni.sh logic :-)  )

tiswanso:
Gate is failing because it's setting up DinD with flannel and the flannel CNI .conflist doesn't have the cniVersion set:

tiswanso:
Looks like kubelet doesn't ignore .conflist files without cniVersion set, only if the cniVersion format is bad.  
https://github.com/containernetworking/cni/blob/master/libcni/conf.go#L77-L84
https://github.com/kubernetes/kubernetes/blob/954996e231074dc7429f7be1256a579bedd8344c/pkg/kubelet/dockershim/network/cni/cni.go#L147

deployments/kubernetes/install/scripts/install-cni.sh
```diff
@@ -20,25 +20,36 @@ function rm_bin_files() {
...
-# default to first file in `ls` output
-# if dir is empty, default to a filename that is not likely to be lexicographically first in the dir
-CNI_CONF_NAME=${CNI_CONF_NAME:-$(ls ${MOUNTED_CNI_NET_DIR} | head -n 1)}
-CNI_CONF_NAME=${CNI_CONF_NAME:-YYY-istio-cni.conflist}
-KUBECFG_FILE_NAME=${KUBECFG_FILE_NAME:-ZZZ-istio-cni-kubeconfig}
-CFGCHECK_INTERVAL=${CFGCHECK_INTERVAL:-1}
+# find_cni_conf_file
+#   Finds the CNI config file in the mounted CNI config dir.
+#   - Follows the same semantics as kubelet
+#     https://github.com/kubernetes/kubernetes/blob/954996e231074dc7429f7be1256a579bedd8344c/pkg/kubelet/dockershim/network/cni/cni.go#L144-L184
```
john-a-joyce: kubelet covers a case where someone put a conflist into a .conf file.   Did you want to cover that case as well or leave for another change if needed.

tiswanso: @john-a-joyce and I double-checked and kubelet isn't allowing putting a .conflist in a .conf file.  The code converts the .conf content to an internal .conflist value and uses it.

PR #98, opened 2019-03-28
bugfix: when pod has more than 2 containers and none of them are isti…

iandyh:
…o-proxy, do not inject iptable rules

As tittle says. 

Thank you.

### DIFF
--- a/cmd/istio-cni/kubernetes.go
+++ b/cmd/istio-cni/kubernetes.go
@@ -55,11 +55,11 @@ func newK8sClient(conf PluginConf, logger *logrus.Entry) (*kubernetes.Clientset,
 
 // getK8sPodInfo returns information of a POD
 func getK8sPodInfo(client *kubernetes.Clientset, podName, podNamespace string) (containers []string,
-	labels map[string]string, annotations map[string]string, ports []string, err error) {
+	labels map[string]string, annotations map[string]string, ports []string, hasSidecar bool, err error) {
 	pod, err := client.CoreV1().Pods(string(podNamespace)).Get(podName, metav1.GetOptions{})
 	logrus.Infof("pod info %+v", pod)
 	if err != nil {
-		return nil, nil, nil, nil, err
+		return nil, nil, nil, nil, false, err
 	}
 
 	containers = make([]string, len(pod.Spec.Containers))
@@ -72,6 +72,7 @@ func getK8sPodInfo(client *kubernetes.Clientset, podName, podNamespace string) (
 
 		if container.Name == "istio-proxy" {
 			// don't include ports from istio-proxy in the redirect ports
+			hasSidecar = true
 			continue
 		}
 		for _, containerPort := range container.Ports {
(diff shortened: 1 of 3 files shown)


tiswanso:
The logic for detecting when to do redirect was changed to be triggered by the annotation: "sidecar.istio.io/inject"

Are you seeing conditions where the istio-proxy container is injected and redirect is required but the annotation is not set?   IMO, don't want to always redirect whenever we see an istio-proxy container and multiple pods.

tiswanso:
Maybe best way to explain is if you could clarify what the issues are with this logic:  https://github.com/istio/cni/blob/master/cmd/istio-cni/main.go#L188-L210

iandyh:
> IMO, don't want to always redirect whenever we see an istio-proxy container and multiple pods.

The current changes does not change the original logic. I only added the sidecar detection. 

> Maybe best way to explain is if you could clarify what the issues are with this logic:

We are seeing cni is setting up the iptable rules even there is no `istio-proxy` present(no istio-injection label set). Because the currently logic will add the redirect whenever the pod has more than 1 container if no annotation is set.

iandyh:
I think the redirect should only be added when there is `istio-proxy` container in the pod.

iandyh:
@tiswanso ping

tiswanso:
@iandyh -- I tried to recreate what you're seeing and am not hitting it.  I see the pod getting excluded if there's not any sidecar annotation--logs from kubelet:
```
Apr 02 18:35:55 bxb-ds-70 kubelet[20724]: 2019-04-02 18:35:55.909 [INFO][21931] <nil> <nil>: Pod helloworld-v3-5f45974c9d-5j5c8 excluded due to not containing sidecar annotation
```

I used this deployment spec to test:
```
apiVersion: v1
kind: Service
metadata:
  name: helloworld
  labels:
    app: helloworld
spec:
  ports:
  - port: 5000
    name: http
  - port: 80
    name: sleep
  selector:
    app: helloworld
---
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: helloworld-v3
  labels:
    version: v3
spec:
  replicas: 1
  template:
    metadata:
      labels:
        app: helloworld
        version: v3
    spec:
      containers:
      - name: helloworld
        image: istio/examples-helloworld-v1
        resources:
          requests:
            cpu: "100m"
        imagePullPolicy: IfNotPresent #Always
        ports:
        - containerPort: 5000
      - name: sleep
        image: pstauffer/curl
        command: ["/bin/sleep", "3650d"]
        imagePullPolicy: IfNotPresent
        ports:
        - containerPort: 80
```

iandyh:
@tiswanso Sorry for the late reply. I've been busy with other testing this week.

I've tried your deployment in our cluster with istio 1.1.1 installed. Here is the condition:
1. When namespace has no label istio-injection, the iptables are still installed, result to external connection being rejected. (we don't have egressgateway) (unexpected)
2. When namespace has no label and annotation contains `sidecar.istio.io/inject: "false"`, iptables are not installed, external connection can be established. (expected)

iandyh:
@tiswanso Hi. After some investigation, we found that in Istio 1.1.0, it actually screwed up the version cni was being used. It was pointing to a 5 month old image which was causing the injection issue.

Current version 1.1.2 has no such issue any more.

The PR was sent was because I was following the logic in the documentation and found that the current implementation differs from the description. I updated the docs with current logic:

https://github.com/istio/cni/pull/101

Thanks for your time.

PR #99, opened 2019-03-28
Add environment variable for cni configuration location.

blairboy362:
In cases where the configuration is not in the default location the
changes the install script makes to existing cni config (such as calico)
point calico's "downstream" config to the wrong place. This should fix
that.


john-a-joyce:
@blairboy362 - you need to be on the google CLA before we can let this merge directly from you

blairboy362:
I signed it!

PR #100, opened 2019-04-09
Incorporate IPv6 changes from istio/istio 11916

tiswanso:
Adds tests using circleCI KDC bringup in ipv6 mode and IPv6 e2e-simple testcase fixes in istio/istio#13753

- istio-iptables.sh changes are identical to istio/istio master with the addition of the ip6tables function to retry on failure to handle potential concurrent ip6tables locking on the same node.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -265,7 +265,7 @@ jobs:
       <<: *dind-env-ipv6
       ISTIO_BRANCH: master
       ISTIO_HUB: docker.io/tiswanso
-      ISTIO_TAG: ipv6-13117
+      ISTIO_TAG: ipv6-13563
       ISTIO_REMOTE: https://github.com/tiswanso/istio.git
       ISTIO_REMOTE_BRANCH: simpletest_ipv6
       CNI_ENABLE: false
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -132,7 +132,7 @@ e2e-dind-test: &e2e-dind-test
     - run:
         name: Run e2e test
         command: |
-          docker run --name=istio-e2e --network=container:kube-master -t -e CNI_TAG=${CIRCLE_SHA1} -e CNI_HUB='docker.io/istio' -e HUB=${ISTIO_HUB:-gcr.io/istio-release} -e TAG=${ISTIO_TAG:-master-latest-daily} -e ISTIO_REMOTE=${ISTIO_REMOTE} -e ISTIO_REMOTE_BRANCH=${ISTIO_REMOTE_BRANCH} -e SKIP_CLEAN='1' e2e-runner:latest /go/run_e2e.sh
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -238,6 +236,18 @@ jobs:
...
     <<: *e2e-dind-test
 
+  e2e-dind-ipv6:
+    <<: *dind-defaults
+    environment:
+      <<: *dind-env-ipv6
+      ISTIO_BRANCH: master
+      ISTIO_HUB: tiswanso
+      ISTIO_TAG: ipv6-13985
+      ISTIO_REMOTE: https://github.com/tiswanso/istio.git
```
john-a-joyce: Nit to consider - Using Istio_remote has the potential to be confused with istio-remote.

.circleci/run_e2e.sh
```diff
@@ -2,21 +2,38 @@
...
-# (the e2e_simple test installs Istio).
-helm template --values ${chartdir}/istio-cni/values.yaml --name=istio-cni --namespace=kube-system --set "excludeNamespaces={}" --set hub=${CNI_HUB} --set tag=${CNI_TAG} --set pullPolicy=IfNotPresent --set logLevel=${CNI_LOGLVL:-debug}  ${chartdir}/istio-cni > istio-cni_install.yaml
-kubectl apply -f istio-cni_install.yaml
+echo "k8s version"
+kubectl version
+
+echo "k8s Nodes"
+kubectl get nodes
+
+if [[ "${CNI_ENABLE}" == "true" ]]; then
```
john-a-joyce: There is now an option to have the test install the CNI for you.  You could possibly have used that but this is also fine.

tools/deb/istio-iptables.sh
```diff
@@ -360,31 +479,144 @@ for internalInterface in ${KUBEVIRT_INTERFACES}; do
...
+  # localhost.
+  ip6tables -t nat -A ISTIO_OUTPUT -d ::1/128 -j RETURN
+
+  # Apply outbound IPv6 exclusions. Must be applied before inclusions.
+  if [ ${#ipv6_ranges_exclude[@]} -gt 0 ]; then
+    for cidr in "${ipv6_ranges_exclude[@]}"; do
+      ip6tables -t nat -A ISTIO_OUTPUT -d "${cidr}" -j RETURN
+    done
+  fi
+  # Apply outbound IPv6 inclusions.
```
john-a-joyce: The IPv4 has a conditional for Kubevirt interfaces here as well as the kubevirt interface within include block below.   You are missing that first kubevirt conditional.   Was that intentional?   Why isn't it needed for the IPv6 case.

.circleci/run_e2e.sh
tiswanso: I think I need to do it this way because I need to use the exact CNI_HUB and CNI_TAG built and copied into the KDC nodes.

tools/deb/istio-iptables.sh
tiswanso: The change other than diffs at L291 - L391 is all identical to the istio/istio master version of the script so that might be mistake in the other as well.  The L291-391 change is just to make ip6tables calls retry if they fail just like we do with iptables calls due to the locking problem.

@sbezverk  --

john-a-joyce:
Is it time to remove the WIP flag and description.

.circleci/run_e2e.sh
john-a-joyce: You should be able to set the CNI_HUB and tag differently on the Istio side.    I recall that someone was arguing in my PR that you don't need different HUB and TAG between Istio and CNI and pretty sure I added the capability for just this type of thing.   It isn't a big deal if you don't want to change this,  but in theory you could let the istio side test infra do this for you.

tiswanso: Forgot to mention that the other reason is to use the istio/cni helm charts from the changeset under test.

PR #101, opened 2019-04-11
docs: align the injection logic with the code

iandyh:


john-a-joyce:
@iandyh - Can you try to get yourself added to the google CLA so we can allow this to merge under your name

README.md
```diff
@@ -278,8 +278,9 @@ Workflow:
 1.  Get k8s pod info
     1.  Determine containerPort list
 1.  Determine if the pod needs to be setup for Istio sidecar proxy
-    1.  If pod has a container named `istio-proxy` AND pod has more than 1 container
+    1.  If pod has more than 1 container
```
john-a-joyce: I don't think this is the best way to work this.   The original text was more a work flow not specifically a map of the logic.   Your change moves more to strict logic map but by doing so its not fully clear what happens to when a pod has only one container.   Maybe  this could be improved by stating the conditions where the pods will have redirect rules
"Pods will have redirect rules when they have more than one container provided neither the sidecar.istio.io/inject annotation is set to false nor is sidecar.istio.io/status annotation missing"

While you are fixing this up you might consider combining bullets 2 & 3.   The code gets the pod information (including a port list) to determine if it needs to be setup for Istio sidecar proxy.

iandyh: > Your change moves more to strict logic map but by doing so its not fully clear what happens to when a pod has only one container. 

The redirection rules setup happens after istio-proxy container being injected, therefore 1 container situation will never be covered here(no redirections at all). This is exactly what the first `if` is doing. Is there any case that one container still needs redirect rules in Istio?

>The code gets the pod information (including a port list) to determine if it needs to be setup for Istio sidecar proxy.

I think the bullet 4 is covering this case. :P

Overall, the reason why I submit this patch is because previous docs was misleading. The actual behaviour from the code is different from the documentation.

iandyh:
@john-a-joyce Thanks for the reply. Yes, I am requesting the CLA within our company.

README.md
john-a-joyce: I definitely agree that the original text is stale and should be fixed up so I am glad you submitted this.   I do think your wording needs to change.  Since this was a work flow you should use the best english to describe what is done vs. just exactly move the code logic to a text format.   The code is always there if someone wants the exact code logic.

iandyh: Hi. I've changed the docs to make it more generic. Thanks for the feedback. 

Please notes: I'll be away for holiday in the next two weeks so will not be able to respond.

PR #102, opened 2019-04-18
[WIP] Experimental sidecar injection from within Istio CNI

luksa:
Implementation of https://github.com/istio/istio/issues/11366

This is an experimental solution for running the sidecar from within the Istio CNI plugin instead of actually injecting it into the Pod definition. 

The CNI plugin determines if the sidecar needs to be started (depending on the pod's annotations; it ignores pods where the sidecar has already been injected by istioctl or the mutating webhook). The sidecar is started through CRI and then managed by the `istio-proxy-agent` container running inside the existing `istio-node` pod deployed by istio-cni. 

I'll re-target the PR to a special feature branch (as discussed in the meeting a few weeks back) as  soon as someone creates it.


loewenstein:
> The sidecar is started through CRI and then managed by the istio-proxy-agent...

So we are partially re-implementing core Kubernetes functionality? I.e. we manage the sidecar container ourselves? How does "normal" scheduling take the additional resources into account that Istio sidecars consume?

tiswanso:
I've pushed the contents of this PR into a new branch `sidecar_management`:  https://github.com/istio/cni/tree/sidecar_management

Please continue use this PR for commentary/discussion.  Continued dev/fixes should go into the new branch.

tiswanso:
@luksa -- do you have alternate build instructions to `make build`?   From the Makefile diffs I think dependencies/init are intended to be covered by `make build`'s dependent targets.  The build is failing in CI for this so might be due to something in your dev env not included by `make depend` -- or maybe golang version dependencies?

tiswanso:
@loewenstein -- Good question.  This topic has come up every time we discuss envoy sidecar proc control moving somewhere else.  I'm not sure if/how @luksa has solved the resource accountability issue but I think there's a decent amount of support for this approach even with the trade-off.  Some folks have thrown some ideas out in the networking wg meetings--e.g. making the sidecar injection add a no-op container with the resources request values for the scheduler.  

But, architecturally, this approach is viewing the Istio proxy as a network function provided by the Istio CNI.  I don't know if k8s networking SIG has a mechanism for accounting for per-pod networking resources.  

Our approach here could be relatively short-lived as projects like Network Service Mesh (NSM) gain traction and add more functionality that could cover this scenario.  I know there's talk of using NSM for integrating network function chaining into k8s networking.

john-a-joyce:
Marko - I am closing this as it is quite stale and you decided to use a different approach.

PR #109, opened 2019-05-23
Move Dockerfile to Ubuntu Xenial and fix linter issues

rlenglet:
Fixes: istio/istio#14320
Signed-off-by: Romain Lenglet <rlenglet@google.com>

### DIFF
--- a/deployments/kubernetes/Dockerfile.install-cni
+++ b/deployments/kubernetes/Dockerfile.install-cni
@@ -1,17 +1,16 @@
-FROM amd64/alpine:3.8
+FROM ubuntu:xenial
 
 LABEL description="Istio CNI plugin installer."
 
-# add bash to allow install-cni.sh to trap process signals
-RUN apk --update add jq bash && \
-    rm -rf /var/lib/apt/lists/* && \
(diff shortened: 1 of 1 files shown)


deployments/kubernetes/Dockerfile.install-cni
```diff
@@ -1,17 +1,16 @@
...
 
 LABEL description="Istio CNI plugin installer."
 
-# add bash to allow install-cni.sh to trap process signals
-RUN apk --update add jq bash && \
-    rm -rf /var/lib/apt/lists/* && \
-    rm /var/cache/apk/*
+RUN apt-get update && apt-get install -qqy --no-install-recommends \
+jq='1.5*' \
+&& rm -rf /var/lib/apt/lists/*
```
venilnoronha: nit: could you reformat this block of code? I'd prefer each line to end in `&& \` like previously.

rlenglet: @venilnoronha I've been following the same style as in https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run. With this style, adding or removing a package only changes a single line.

PR #113, opened 2019-05-28
Update istio-iptables.sh

rlenglet:
Re-sync `istio-iptables.sh` with istio/istio. Re-apply local changes to it (f63b5369ff444a8f252d860a8bdc0107a121c749) and fix ShellCheck issues. Fix the local changes to use `return` instead of `exit` in the `iptables` and `ip6tables` functions.


john-a-joyce:
This change and https://github.com/istio/cni/pull/100 have much of the same content.   What is the motivation for pushing this as a separate PR versus adding to or appending to PR #100 when it merges.

rlenglet:
@john-a-joyce 

> This change and #100 have much of the same content. What is the motivation for pushing this as a separate PR versus adding to or appending to PR #100 when it merges.

That is not true. I have split #100. 3 commits moved into this PR, and 3 commits remained in #100.

rlenglet:
@john-a-joyce

> This change and #100 have much of the same content. What is the motivation for pushing this as a separate PR versus adding to or appending to PR #100 when it merges.

Oh my bad, I didn't look at #100.

john-a-joyce:
I don't want to argue about little details at the macro level there is much overlap so I am trying to understand the motivation and why we shouldn't let 100 merge first or fix it up if there are issues with it.   Are you doing this mostly for testing?

rlenglet:
This is passing all tests. Please review!

tiswanso:
We should hold this for #100 which is ready to go IMHO.  I like the approach you have here of updating the istio-iptables.sh with our CNI diffs until we can get a common go version of istio-iptables.sh.  The overall diffs here look good but I do have some comments.  Reviewing & will add.

.circleci/config.yml
```diff
@@ -148,10 +148,12 @@ e2e-dind-test: &e2e-dind-test
         path: /tmp/cluster_state
     - run:
         name: Bring down the cluster
+        when: always
```
tiswanso: I don't like this for "rerun with ssh" usage of circle for debugging.  I don't think there's any harm in leaving the KDC clusters up since they're on the local workflow's remote_docker instance, which get cleaned up by circleci on job termination, AFAIK.

tools/packaging/common/iptables-retry.diff
```diff
@@ -0,0 +1,108 @@
...
++        else
++            return $err_code
++        fi
++    done
++    set -o errexit
++}
++#
++# Function wrapping ip6tables to retry in case of failure if IPTABLES_RETRY is "true".
++#
++function ip6tables {
```
tiswanso: see how I did this in #100.  I prefer that approach to consistent retry logic for both ipv4/v6.  Also, the ip6tables cleanup needs to be before this and, IMHO, should be unconditional like the v4 chains cleanup.

.circleci/config.yml
rlenglet: @tiswanso OK, I'll remove that. I just found that surprising by reading the list of steps. How about moving the `store_artifacts` step to the end of the list, like it's done in all istio/istio jobs, for consistency?

tools/packaging/common/iptables-retry.diff
rlenglet: @tiswanso Agreed. Just merge #100 and I'll update to pick that up.

.circleci/config.yml
rlenglet: Actually, I'll just remove the cleanup steps. istio/istio integration tests don't clean up either.

costinm:
Yes, cleanup is waste of time - the entire VM is wiped after the test, and
if you want to debug failures with ssh you want the env to be preserved.

One suggestion: maybe would be good to drop the tproxy stuff that
complicates the script, and leave it for the go implementation. I would
also remove the bash-specific code and validation ( added recently and
again complicated and wouldn't help ). In separate PRs, of course.

tools/packaging/common/iptables-retry.diff
rlenglet: @tiswanso I picked up the change to unconditionally cleanup the `ip6tables` tables. Please review.

rlenglet:
Is there anything blocking merging this PR?

john-a-joyce:
@rlenglet - I think the only thing was to let PR 100 go in first.    I just approved that and it auto-merged

john-a-joyce:
So if you resolve the conflicts then we can merge this.

tools/packaging/common/istio-iptables.sh
```diff
@@ -116,6 +116,64 @@ function isIPv6() {
...
+IPTABLES_CMD=$(which iptables)
+#
+# The path to the ip6tables command.
+#
+# shellcheck disable=SC2230
+IP6TABLES_CMD=$(which ip6tables)
+#
+# If true, retries iptables & ip6tables commands in case they fails.
+#
+IPTABLES_RETRY=false
```
tiswanso: Ahh... I see what you did there.  Took me a min.  Hopefully with the tighter change block future patch applies will be smooth.

rlenglet: Yeah, that change was necessary to make shellcheck pass. It doesn't like what looks like using a function before it's defined.

PR #117, opened 2019-06-12
excludeOutboundPorts to facilitate headless data stores

nisan270390:
Currently inbound ports can be excluded from redirection, and outbound IP addresses can be excluded from redirection, but not outbound ports. This fills out the support for that, by adding an traffic.sidecar.istio.io/excludeOutboundPorts annotation.

Istio has the ability to exclude outbound ports by merging #13970 pull request, and this  currently PR is adding the exclude outbound ports to istio cni as well

### DIFF
--- a/cmd/istio-cni/redirect.go
+++ b/cmd/istio-cni/redirect.go
@@ -37,10 +37,11 @@ const (
 	defaultRedirectExcludePort   = defaultProxyStatusPort
 	defaultKubevirtInterfaces    = ""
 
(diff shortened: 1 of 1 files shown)


tiswanso:
@john-a-joyce fyi

john-a-joyce:
Note:  https://github.com/istio/istio/pull/14171 is needed to allow this to work if the annotations aren't  directly added to the pod spec.

PR #118, opened 2019-06-12
Make file modifications atomic in install-cni.sh

rlenglet:
This eliminates a race condition where CNI plugins and kubelet could read garbage when starting new pods while the CNI files are getting updated.


tiswanso:
Still need to review in detail, but has an issue been seen with this?  either in real deployments or in integration testing?

rlenglet:
> Still need to review in detail, but has an issue been seen with this? either in real deployments or in integration testing?

This is fixing a race condition between the file updates and container creations. This has not been reported yet since it should be a rare occurrence. But it's a real bug and I'm fixing it proactively.

What are your concerns?

rlenglet:
Maybe one reason why this has not been found is that there are no tests installing the CNI while pods are getting created?

tiswanso:
> > Still need to review in detail, but has an issue been seen with this? either in real deployments or in integration testing?
> 
> This is fixing a race condition between the file updates and container creations. This has not been reported yet since it should be a rare occurrence. But it's a real bug and I'm fixing it proactively.
> 
> What are your concerns?

It's more about the priority for getting this in.

rlenglet:
Now that 1.2 has been released, please review this again? Thanks!

rlenglet:
Not stale. Still waiting for review for more than 4 months.

PR #123, opened 2019-06-13
Backport #122 "Use consistent image tag in Helm values"

rlenglet:
Fixes: https://github.com/istio/istio/issues/14805


rlenglet:
> We might want to make this release-1.2-latest-daily (if that is even built?). It doesn't matter much though since the release will override the tag anyways

@howardjohn I've followed istio/istio, which still uses `master-latest-daily` in `release-1.2`. I was thinking we'll update that at release time before tagging the release.

howardjohn:
Oh maybe we need to change that there. Its not really used for anything though except if a dev installs from the repo. Tests and release both don't use it, so not a big deal

rlenglet:
@howardjohn ok, sent out https://github.com/istio/istio/pull/14813

PR #125, opened 2019-06-13
Update default tag in Helm charts for 1.2

rlenglet:
Keep consistent with istio/istio, cf. https://github.com/istio/istio/pull/14813

### DIFF
--- a/deployments/kubernetes/install/helm/istio-cni/values.yaml
+++ b/deployments/kubernetes/install/helm/istio-cni/values.yaml
@@ -1,5 +1,5 @@
 hub: gcr.io/istio-release
-tag: master-latest-daily
+tag: release-1.2-latest-daily
 pullPolicy: Always
 
 logLevel: info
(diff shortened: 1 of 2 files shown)


tiswanso:
@sdake -- this is a sample override file for gke (`values_gke.yaml`) so we don't need as much caution here, IMO.  Probably, we can just remove those settings and any other copied values from the file.  No tests or demos use the file at this point.

deployments/kubernetes/install/helm/istio-cni/values.yaml
```diff
@@ -1,5 +1,5 @@
 hub: gcr.io/istio-release
-tag: master-latest-daily
+tag: release-1.2-latest-daily
```
sdake: @tiswanso this is not a gke override -this is the main config.  The test-infra seds (iirc) master-latest-daily to produce the version number.

tiswanso: yeah, my bad... there was a similar fix yesterday on master for the gke file only and I assumed this was a cherrypick.

tiswanso:
@rlenglet -- see this PR from @utka for what @sdake is referring to for CNI:  https://github.com/istio/istio/pull/14806/files

tiswanso:
@sdake @rlenglet -- the sed in istio/istio#14806 should still work with the diffs in this PR.  It matches on `tag: .*-latest-daily`

sdake:
Thanks @tiswanso for hunting that down.  I was looking for it :)

PR #119, opened 2019-06-12
Merge latest master changes to release-1.2

tiswanso:
```
d93746cb excludeOutboundPorts to facilitate headless data stores (#117) 
93e77bf6 Update istio-iptables.sh (#113)
8a6681fb Incorporate IPv6 changes from istio/istio 11916 (#100)
154bcfea Update common files. (#115)
971dca04 Update common files (#114)
3b806e0c Add all CircleCI jobs as preconditions to Mergify success (#112)
270ca989 Update common files and fix lint warnings (#110)
```

PR #127, opened 2019-06-17
Update common files.

geeknoid:


rlenglet:
Should this be backported to 1.2 as well?

geeknoid:
Nope, no need for backporting.

tiswanso:
fyi, it's already backported with #119.  #119 sync'd 1.2 with master as of 6/13/2019

PR #129, opened 2019-06-19
Migrate from dep to Go modules

rlenglet:
Move from Go deps to Go modules, using `go mod` commands:
```
go mod init
go mod vendor
go mod tidy
```

Keep the `vendor/` directory to speed up builds since the build is not configured to use modules, as is done in istio/istio, cf. https://github.com/istio/istio/pull/13899.

Fixes https://github.com/istio/istio/issues/14431


ericvn:
Just confirming you did the `go mod vendor` after the `go mod tidy`. I assume you did, but you had them in the inverse order in the comment.

PR #130, opened 2019-06-21
Remove outdated istio-cni*.yaml manifests

rlenglet:
The `.../install/manifests/*.yaml` are outdated:

- they use unreleased images
- they use different resource names and labels from the Helm charts
- they are not tested
- they are not referenced anywhere in the Istio documentation, which only uses Helm

Delete them and remove any reference in `README.md`.


README.md
```diff
@@ -213,8 +215,8 @@ $ gcloud logging read "resource.type=gce_instance AND jsonPayload.SYSLOG_IDENTIF
 
 ### Overview
 
-- [istio-cni.yaml](deployments/kubernetes/install/manifests/istio-cni.yaml)
-   - manifest for deploying `install-cni` container as daemonset
+- [istio-cni.yaml](deployments/kubernetes/install/helm/istio-cni/templates/istio-cni.yaml)
```
costinm: Please move the templates to istio/installer - that's where all other install templates are maintained.

rlenglet: That's out of the scope of this PR.

costinm:
Well - if you are changing the README on installation, add a TODO or
mention that this is just a temporary path and will change.

costinm:
I'm not suggesting to do the move in this PR, of course - it's not even
possible, different repo :-)

Just a TODO or not in the README is sufficient for this PR.

rlenglet:
> the README and templates should be in istio/install.

> this is just a temporary path and will change.

This is not a temporary location. This is where the release script is currently pulling the `istio-cni` chart from to be included into releases, cf. https://github.com/istio/istio/commit/4b80960677bddbdee7e8f7f6344c231e128b18cb.

I don't know what the plan is re: integration with the installer, but changing this would change how the release is built.

Definitely out of the scope of this PR.

PR #141, opened 2019-08-14
Make use of RBAC v1

johscheuer:
This PR changes the RBAC API version to the current v1 (which is GA since I think 1.8: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.8.md#auth).


howardjohn:
/ok-to-test

sorry the ci got broken by a dependency, will merge this in when it's fixed , someone is working on it

PR #142, opened 2019-08-14
Remove vendor

howardjohn:


howardjohn:
broken until we have golang 1.12

tiswanso:
The problem is this line in .circleci/config.yaml:

PR #147, opened 2019-08-19
Support ipv6 in e2e tests

howardjohn:


howardjohn:
We would need a dedicated ipv6 node pol for this

PR #149, opened 2019-08-19
Remove circle steps covered by prow

howardjohn:


howardjohn:
Will fail until https://github.com/istio/test-infra/pull/1693 merges

tiswanso:
Sorry, I haven't kept up with the test-infra changes.  So, now with the PRs you've merged istio/test-infra is able to trigger on istio/cni PRs; build istio/cni binaries and install-cni; and trigger e2e tests referring to the cni images under test and a version of istio/istio (head or some tagged version)?

howardjohn:
Exactly. The only thing missing is ipv6 support on prow. TBD on solution to
that

tiswanso:
re: IPv6 support in prow... easiest would likely be to get KinD IPv6 to work.  I'm not sure the details of IPv6 on KinD.  Last I looked a few months ago it wasn't ready but looks like it's available as of KinD 0.4.  
https://kind.sigs.k8s.io/docs/user/quick-start#ipv6-clusters

howardjohn:
Yeah we can run KinD ipv6 but the docker container needs to have ipv6
support, which we don't have in prow (for now?) as far as I know.
Kubernetes does ipv6 in kind but looks like they have some custom nodes

PR #172, opened 2019-09-13
Disable retrying ip6tables cleanup when IPv6 is disabled

rlenglet:
Fixes https://github.com/istio/istio/issues/15895


lambdai:
Does go version iptables exist? Do you need to sync with go version iptables?

rlenglet:
> Does go version iptables exist? Do you need to sync with go version iptables?

No, only this shell script needs to be updated.

lambdai:
As a global variable IPTABLES_RETRY mutated 3 times. I don't have a easy view when I see each iptables/ip6tables is retrying or not
Consider this iptables script is not changing it's probably fine for now.

PR #175, opened 2019-09-18
Add the possibility to create RoleBinding for PSP

johscheuer:
Backport of: https://github.com/istio/installer/pull/396

### DIFF
--- a/deployments/kubernetes/install/helm/istio-cni/templates/istio-cni.yaml
+++ b/deployments/kubernetes/install/helm/istio-cni/templates/istio-cni.yaml
@@ -146,7 +146,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
 
 ---
-{{- if .Values.psp_cluster_role }}
+{{- if hasKey .Values "psp_cluster_role" }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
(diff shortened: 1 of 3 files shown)


johscheuer:
With my local helm version everything works fine:

```bash
$ helm version --client
Client: &version.Version{SemVer:"v2.14.3", GitCommit:"0e7f3b6637f7af8fcfddb3d2941fcc7cbebb0085", GitTreeState:"clean"}
$ helm lint deployments/kubernetes/install/helm/istio-cni
==> Linting deployments/kubernetes/install/helm/istio-cni
Lint OK

1 chart(s) linted, no failures
```

deployments/kubernetes/install/helm/istio-cni/templates/istio-cni.yaml
```diff
@@ -144,3 +144,20 @@ kind: ServiceAccount
 metadata:
   name: istio-cni
   namespace: {{ .Release.Namespace }}
+
+---
+{{- if hasKey .Values "psp_cluster_role" }}
```
rlenglet: Instead of `hasKey`, could you instead test for string emptyness, and add this into the `values*.yaml` files for documentation? The `README.md` file doesn't ship into the Istio release, so that documentation is lost. Adding that into `values*.yaml` with a comment would be very useful.

rlenglet:
@johscheuer Also, I would really appreciate if you could submit a PR to document the new parameter in the CNI docs, to be consistent: https://istio.io/docs/setup/additional-setup/cni/.
Thanks again, this is very useful!

deployments/kubernetes/install/helm/istio-cni/templates/istio-cni.yaml
johscheuer: I do !

PR #178, opened 2019-09-19
Update common files and fix a linter warning.

geeknoid:


geeknoid:
Depends on https://github.com/istio/test-infra/pull/1826 to pass lintmodern_cni.

common/config/.golangci.yml
```diff
@@ -42,6 +42,8 @@ linters:
     - nakedret
     - prealloc
     - scopelint
+    - funlen
```
rlenglet: I commented those two linters in `release-1.3` branch. They were failing like here.

rlenglet: It seems like https://github.com/istio/test-infra/pull/1826 is only fixing `master`, so I'll keep those commented out in `release-1.3` I suppose?

geeknoid: Yes.

In general, I think it's better to not aggressively merge things from master into a release branch. Tool chain updates (which is what 'make update-common' effectively is) should definitely not be copied to release branches. Only targeted specific fixes should be moved over if at all possible, in order to minimize the delta of any patch releases made from the branch.

PR #184, opened 2019-09-30
Update common files.

geeknoid:


geeknoid:
Needs istio/test-infra#1861

PR #186, opened 2019-10-02
Use su-sxec to chown /home within build container

sdake:
The home directory is owned by root.  When running in a non-root
context, Kind will be unable to write to the /home directory.  It
is necessary to chown the directory to $(UID) but only when
BUILD_WITH_CONTAINER=1 is true.

Depends-On: https://github.com/istio/common-files/pull/91


sdake:
@howardjohn you may find this PR helpful in your work on istio/installer porting.  Check for the logic around su-exec and adding su-exec to the makefile prior to the creation of the ~/.kube/config file.

Makefile
```diff
@@ -69,6 +69,7 @@ $(info Building with the build container: $(IMG).)
 TIMEZONE=`readlink $(READLINK_FLAGS) /etc/localtime | sed -e 's/^.*zoneinfo\///'`
 
 RUN = $(CONTAINER_CLI) run -t -i --sig-proxy=true -u $(UID):docker --rm \
+	-e BUILD_WITH_CONTAINER="$(BUILD_WITH_CONTAINER)" \
```
sdake: update to IN_BUILD_CONTAINER

Makefile.core.mk
```diff
@@ -184,7 +184,12 @@ $(ISTIO_OUT) $(ISTIO_BIN):
 precommit: lint_modern fmt_modern
 
 prow-e2e:
-	./test/prow-e2e.sh
+# Needed as the volume mount /home in the container is mounted as UID 0.
+# Kind needs to write to /home, so we chown it to our UID.
+ifeq ($(BUILD_WITH_CONTAINER),1)
+	@/usr/local/bin/su-exec 0:0 chown $(shell id -u) /home
```
sdake: su-exec should be in the path now.

sdake:
waiting for prow to churn out a new build container until removing WIP.

PR #187, opened 2019-10-02
Expunge circleci from cni repo

sdake:
Removing circleci. No longer necessary for CNI testing. If
someone steps up to maintain IPv6 support and IPv6 support cannot
be implemented in PROW, CircleCI may be brought back into the
implementation. Until that happens, this check is just failing and
there is nothing worse than code that doesn't work ;)


sdake:
@geeknoid would you be kind enough to remove the circleci build integration?  I believe this can only be done by github administrators.

Cheers
-steve

sdake:
> @geeknoid would you be kind enough to remove the circleci build integration? I believe this can only be done by github administrators.
> 
> Cheers
> -steve

Nevermind, I sorted this problem out on my own.

Chers
-steve

PR #188, opened 2019-10-02
Place istio/cni on the BUILD_WITH_CONTAINER=1 plan

sdake:
Currently prow-e2e doesn't work well with BUILD_WITH_CONTAINER=1
however, we are very close to having all make targts work with
BUILD_WITH_CONTAINER=1.

### DIFF
--- a/Makefile.overrides.mk
+++ b/Makefile.overrides.mk
@@ -12,5 +12,5 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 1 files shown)


sdake:
@howardjohn - it appears e2e-prow (which uses docker) is a very slightly flakey.  This is the first time I've seen this test flake.  I know you're doing work on KIND as well with istio/installer.  If you have any thoughts as to the flaky behavior, I would appreciate any feedback you have.

Cheers
-steve

PR #190, opened 2019-10-02
Enable the build container by default.

geeknoid:
- Rename some targets for consistent with other repos.


geeknoid:
Thanks @sdake for all the work to make this possible.

PR #191, opened 2019-10-05
Enable `make e2e`

sdake:
make prow-e2e works in prow, but not locally. This PR enables
e2e to work locally.


sdake:
using https://github.com/istio/cni/pull/192 in place of this PR.

PR #194, opened 2019-10-07
Make `test` work properly

sdake:
Sadly a host mount must be used to share files between the build container
and the test case. The test case uses /tmp for this. This test case will
spew a bunch of backtraces, which it is supposed to spew.  The actual
test results pass.


sdake:
Related-To: https://github.com/istio/tools/pull/417

tiswanso:
I suggest adding `UID ?= $(shell id -u)` to this patch and a `test/README.md` update to allow for OSX users to run this via `sudo UID=0 make test`.   OR we should modify instructions to say to use `BUILD_WITH_CONTAINER=0 make test` in local envs.

PR #197, opened 2019-10-08
Delete an unused script.

geeknoid:


geeknoid:
/retest

One day, it'll pass...

PR #206, opened 2019-10-23
Allow to set custom annotations on pod level

djablonski-moia:
This PR allows to add custom annotations on the Pod level for the DaemonSet. This is beneficial to support use cases where for example custom validators or operators utilize annotations to  check if something should be allowed or not.
As a concrete example, we would like to be able to document that the pods as exceptionally getting broader access rights.

### DIFF
--- a/deployments/kubernetes/install/helm/istio-cni/templates/istio-cni.yaml
+++ b/deployments/kubernetes/install/helm/istio-cni/templates/istio-cni.yaml
@@ -85,9 +85,9 @@ spec:
         # priority scheduling and that its resources are reserved
         # if it ever gets evicted.
         scheduler.alpha.kubernetes.io/critical-pod: ''
-{{- range $key, $value := .Values.cniCustomPodAnnotations }}
-        {{ $key }}: {{ $value | quote }}
-{{- end }}
+        {{- if .Values.podAnnotations }}
+{{ toYaml .Values.podAnnotations | indent 8 }}
+        {{- end }}
     spec:
       nodeSelector:
         beta.kubernetes.io/os: linux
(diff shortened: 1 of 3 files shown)


djablonski-moia:
@googlebot I signed it!

deployments/kubernetes/install/helm/istio-cni/templates/istio-cni.yaml
```diff
@@ -85,6 +85,9 @@ spec:
         # priority scheduling and that its resources are reserved
         # if it ever gets evicted.
         scheduler.alpha.kubernetes.io/critical-pod: ''
+{{- range $key, $value := .Values.cniCustomPodAnnotations }}
```
rlenglet: Follow the same conventions as Istio's other charts, e.g. [Pilot's chart](https://github.com/istio/istio/blob/master/install/kubernetes/helm/istio/charts/pilot/templates/deployment.yaml#L38-L40):
```
         {{- if .Values.podAnnotations }}
{{ toYaml .Values.podAnnotations | indent 8 }}
        {{- end }}
```

Also use the same name `podAnnotations` for consistency.

rlenglet: Also, all Helm values must be documented with their default values in the `values*.yaml` files, preferably with a comment explaining the semantics of the parameter. E.g. for [Pilot's chart](https://github.com/istio/istio/blob/master/install/kubernetes/helm/istio/charts/pilot/values.yaml#L30):
```
podAnnotations: {}
```

rlenglet:
The `lint_cni` test failed because the new parameter must be defined in `values.yaml` and `values_gke.yaml`, as I commented above. This is not optional.

djablonski-moia:
Thanks for the feedback @rlenglet!

I adopted the PR according to your requested changes. Please let me know if there are other improvements to be made.

djablonski-moia:
PR for docs created: https://github.com/istio/istio.io/pull/5257

PR #218, opened 2019-11-14
Exclude pod if istio-init container is present

abhide:
Issue: https://github.com/istio/istio/issues/18021

### DIFF
--- a/cmd/istio-cni/kubernetes.go
+++ b/cmd/istio-cni/kubernetes.go
@@ -57,13 +57,17 @@ func newK8sClient(conf PluginConf) (*kubernetes.Clientset, error) {
 
 // getK8sPodInfo returns information of a POD
 func getK8sPodInfo(client *kubernetes.Clientset, podName, podNamespace string) (containers []string,
-	labels map[string]string, annotations map[string]string, ports []string, err error) {
+	initContainers map[string]struct{}, labels map[string]string, annotations map[string]string, ports []string, err error) {
 	pod, err := client.CoreV1().Pods(podNamespace).Get(podName, metav1.GetOptions{})
 	log.Infof("pod info %+v", pod)
 	if err != nil {
-		return nil, nil, nil, nil, err
+		return nil, nil, nil, nil, nil, err
 	}
 
+	initContainers = map[string]struct{}{}
+	for _, initContainer := range pod.Spec.InitContainers {
+		initContainers[initContainer.Name] = struct{}{}
+	}
 	containers = make([]string, len(pod.Spec.Containers))
 	for containerIdx, container := range pod.Spec.Containers {
 		log.Debug("Inspecting container",
@@ -89,5 +93,5 @@ func getK8sPodInfo(client *kubernetes.Clientset, podName, podNamespace string) (
 		}
 	}
 
-	return containers, pod.Labels, pod.Annotations, ports, nil
+	return containers, initContainers, pod.Labels, pod.Annotations, ports, nil
 }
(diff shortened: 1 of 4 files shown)


cmd/istio-cni/kubernetes.go
```diff
@@ -91,3 +92,17 @@ func getK8sPodInfo(client *kubernetes.Clientset, podName, podNamespace string) (
 
 	return containers, pod.Labels, pod.Annotations, ports, nil
 }
+
+// getK8sInitContainers returns init containers of a POD
+func getK8sPodInitContainers(client *kubernetes.Clientset, podName, podNamespace string)(
+	initContainers map[string]struct{}, err error) {
+	pod, err := client.CoreV1().Pods(podNamespace).Get(podName, metav1.GetOptions{})
```
rlenglet: Avoid doing another k8s API call.
Instead, modify the function above to return the info we need.

cmd/istio-cni/main.go
```diff
@@ -161,13 +163,22 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+			return err
+		}
+		log.Debug("Created Kubernetes client",
+			zap.Reflect("client", client))
+		// Check if istio-init container is present; in that case exclude pod
+		initContainersMap, err := getKubePodInitContainers(client, string(k8sArgs.K8S_POD_NAME), string(k8sArgs.K8S_POD_NAMESPACE))
+		if err != nil {
+			log.Warnf("Error getting Pod data %v", err)
+		}
+		if _, present := initContainersMap[ISTIOINIT]; present {
```
rlenglet: 👍 This is the right logic.

cmd/istio-cni/kubernetes.go
abhide: Done

cmd/istio-cni/main.go
```diff
@@ -166,12 +168,18 @@ func cmdAdd(args *skel.CmdArgs) error {
...
-			containers, _, annotations, ports, k8sErr := getKubePodInfo(client, string(k8sArgs.K8S_POD_NAME), string(k8sArgs.K8S_POD_NAMESPACE))
+			log.Debug("Created Kubernetes client", zap.Reflect("client", client))
+			containers, initContainersMap, _, annotations, ports, k8sErr := getKubePodInfo(client, string(k8sArgs.K8S_POD_NAME), string(k8sArgs.K8S_POD_NAMESPACE))
 			if k8sErr != nil {
 				log.Warnf("Error getting Pod data %v", k8sErr)
 			}
+
+			// Check if istio-init container is present; in that case exclude pod
+			if _, present := initContainersMap[ISTIOINIT]; present {
+				log.Infof("Found istio-init container for pod: %s; namespace: %s", string(k8sArgs.K8S_POD_NAME), string(k8sArgs.K8S_POD_NAMESPACE))
```
rlenglet: Make it clear in the log that the pod is excluded, following a pattern similar to the one below:
```suggestion
				log.Info("Pod excluded due to being already injected with istio-init container",
					zap.String("pod", k8sArgs.K8S_POD_NAME),
					zap.String("namespace", k8sArgs.K8S_POD_NAMESPACE))
```

abhide: done

rlenglet:
@abhide it would be nice to have a similar change in `release-1.4`. This can't be cherry-picked automatically, because I believe there are significant changes in `master`, incl. a change of logging libraries, etc.

abhide:
Thanks @rlenglet I will cherry-pick; fix diffs and post a PR for release-1.4

PR #224, opened 2019-12-12
Sync istio-iptables.sh from istio/istio

abhide:


abhide:
> Thanks!
> I assume you double checked that you pulled the last state of istio-iptables.sh before you deleted it from istio/istio.

yes

PR #225, opened 2019-12-17
Donot redirect app calls back to itself by podip via envoy

hzxuzhonghu:
This is sync from https://github.com/istio/istio/pull/19496


hzxuzhonghu:
@istio/wg-networking-maintainers @abhide  ptal

PR #226, opened 2019-12-17
Adding CNI Repair Daemonset

stewartbutler:
This PR will implement the CNI repair daemonset as detailed in https://tinyurl.com/istio-cni-race. The code is still being written, this PR is just a placeholder for tracking progress and providing a place for implementation discussion.

This PR will not include the race condition detection code; at present, the plan is for that to be included in a separate PR by @lambdai.


cmd/istio-cni-daemonset/main.go
```diff
@@ -0,0 +1,363 @@
...
+	clientSet, err := clientSetup(options)
+	if err != nil {
+		log.Fatalf("Could not construct clientSet: %s", err)
+	}
+	podFixer := brokenPodReconciler{client: *clientSet}
+
+	if options.RunAsDaemon {
+		var wg sync.WaitGroup
+		wg.Add(1)
+		go func(wg sync.WaitGroup) {
```
howardjohn: It seems like this is the same as not running in a goroutine, am I missing something?

cmd/istio-cni-daemonset/main.go
```diff
@@ -0,0 +1,363 @@
...
+// This func checks the state of a container in a pod and returns true if the container
+// matches the defined filter criteria
+func checkTerminationState(filter Filters, state v1.ContainerStateTerminated) bool {
+	// If we are filtering on init container termination message and the termination message of 'state' does not match, exit
+	if s := strings.TrimSpace(filter.InitContainerTerminationMessage); s != "" && s != strings.TrimSpace(state.Message) {
+		return false
+	}
+
+	// If we are filtering on init container exit code and the termination message does not match, exit
+	if ec := filter.InitContainerExitCode; ec != 0 && ec != int(state.ExitCode) {
```
mandarjog: why the extra variable ?

mandarjog:
```go
return filter.InitContainerExitCode == int(state.ExitCode)
```

cmd/istio-cni-daemonset/main.go
```diff
@@ -0,0 +1,363 @@
+// Copyright 2019 Istio Authors
```
mandarjog: nit?:
This is a command line, could be used from istioctl too.
Can you move all the functional code to `pkg` and just *use* that code from here?
There is nothing in the main code that says daemon ...

cmd/istio-cni-daemonset/main.go
stewartbutler: Ah, yes -- leftover code from a refactor. At one point I had both the delete and label code running in separate goroutines, I moved them out into the single reconciliation function and forgot to remove the goroutine boilerplate. Will remove.

cmd/istio-cni-daemonset/main.go
stewartbutler: Yep, I can do so. I was just keeping it in a single file for simplicity, but that'd probably be a better way to architect it.

cmd/istio-cni-daemonset/main.go
stewartbutler: filter.InitContainerExitCode defaults to zero if unset. We never want to flag a pod as broken if the init container returns  an exit code of zero (successful execution). Without checking for non-zero exit code, that statement would end up returning false positives for pods that have successfully completed their init container and are now running their application payload.

I could change the default ExitCode to something nonsensical, I just left it as zero since that is not an exit code we would ever expect to filter on and so is an easy code to check for. 

As for why I used the 'ec' variable, it was mostly just to shorten the if statement and have it match the pattern set by the InitContainerTerminationMessage if statement above it.

pkg/repair/repair.go
```diff
@@ -0,0 +1,220 @@
...
+	// Get a list of all broken pods
+	podList, err := bpr.ListBrokenPods()
+	if err != nil {
+		return
+	}
+
+	for _, pod := range podList.Items {
+		labels := pod.GetLabels()
+		if labels == nil {
+			labels = map[string]string{}
```
mandarjog: nit: labels init can be moved to "else - label not found section"

pkg/repair/repair.go
```diff
@@ -0,0 +1,220 @@
...
+		}
+		if _, ok := labels[bpr.Options.PodLabelKey]; ok {
+			log.Infof("Pod %s/%s already has label with key %s, skipping", pod.Namespace, pod.Name, bpr.Options.PodLabelKey)
+			continue
+		} else {
+			log.Infof("Labeling pod %s/%s with label %s=%s", pod.Namespace, pod.Name, bpr.Options.PodLabelKey, bpr.Options.PodLabelValue)
```
mandarjog: nit: `else` can be removed to unindent code.

pkg/repair/repair.go
```diff
@@ -0,0 +1,220 @@
...
+	// Get a list of all broken pods
+	podList, err := bpr.ListBrokenPods()
+	if err != nil {
+		return err
+	}
+
+	for _, pod := range podList.Items {
+		log.Infof("Deleting broken pod: %s/%s", pod.Namespace, pod.Name)
+		if err := bpr.client.CoreV1().Pods(pod.Namespace).Delete(pod.Name, nil); err != nil {
+			return err
```
mandarjog: Should not terminate loop on the first error. May return a list of errors.

pkg/repair/repair.go
stewartbutler: Fixed

pkg/repair/repair.go
stewartbutler: Fixed

pkg/repair/repair.go
stewartbutler: Fixed it by coalescing all the errors after the loop. Can you let me know if that isn't the approach you had in mind?

stewartbutler:
Just so that there is a link, the problem detection portion of this PR is added to istio-init in https://github.com/istio/istio/pull/19534

stewartbutler:
I think this is ready to go.

cmd/istio-cni-daemonset/main.go
```diff
@@ -0,0 +1,240 @@
...
+	pflag.String(
+		"init-container-termination-message",
+		"",
+		"The expected termination message for the init container when crash-looping because of CNI misconfiguration")
+	pflag.Int(
+		"init-container-exit-code",
+		126,
+		"Expected exit code for the init container when crash-looping because of CNI misconfiguration")
+
+	pflag.StringSlice("label-selectors", []string{}, "A set of label selectors in label=value format that will be added to the pod list filters")
```
rlenglet: If you use the standard Kubernetes label selector API, a single selector should be sufficient as it supports a conjunction of predicates on labels already.
So this should be a single string, and the flag name should be singular.

cmd/istio-cni-daemonset/main.go
howardjohn: We already have a CNI daemonset, are users expected to run a second daemonset as well? maybe I missed context here but it seems like we should just have one daemonset that does both functionality

cmd/istio-cni-daemonset/main.go
```diff
@@ -0,0 +1,240 @@
...
+	if err := viper.BindPFlags(pflag.CommandLine); err != nil {
+		log.Fatal("Error parsing command line args: %+v")
+	}
+
+	if viper.GetBool("help") {
+		pflag.Usage()
+		os.Exit(0)
+	}
+
+	viper.SetEnvPrefix("icnid")
```
howardjohn: what is icdni?

deployments/kubernetes/Dockerfile.istio-cni-daemonset
```diff
@@ -0,0 +1,14 @@
+FROM ubuntu:xenial
```
howardjohn: use :bionic to match other istio components. Or distroless

deployments/kubernetes/cni-daemonset/istio-cni-daemonset.yaml
```diff
@@ -0,0 +1,49 @@
...
+      labels:
+        k8s-app: istio-cni-daemonset
+    spec:
+      serviceAccountName: istio-cni-daemonset-sa
+      tolerations:
+      # Tolerate all taints; we want to run on all nodes
+      - operator: "Exists"
+      containers:
+      - name: istio-cni-daemonset
+        image: gcr.io/istio-release:master-latest-daily
```
howardjohn: this is not valid and master-latest-daily is dead

deployments/kubernetes/cni-daemonset/istio-cni-daemonset.yaml
```diff
@@ -0,0 +1,49 @@
+apiVersion: apps/v1
+kind: DaemonSet
```
howardjohn: I don't see why this is a Daemonset? There is nothing that needs to be on the same node as far as I can tell, we are using standard k8s APIs

cmd/istio-cni-daemonset/main.go
mandarjog: This should be run as (part of the ) cni daemonset or it can be run as a command line utility.

deployments/kubernetes/cni-daemonset/istio-cni-daemonset.yaml
mandarjog: Repairer can run as a command line utility or as a daemon.
Since we already have cni daemonset it should run as a part of it.

cmd/istio-cni-daemonset/main.go
stewartbutler: Shortening of `istio-cni-daemonset` so that you can set ENV variables instead of passing command line arguments, e.g. '--node-name=Test' could be set by using the env variable 'ICNID_NODE-NAME=Test'

deployments/kubernetes/cni-daemonset/istio-cni-daemonset.yaml
stewartbutler: I don't have an issue with having this as a single controller, but I think there was some talk about using this as a general purpose daemonset and adding additional functionality later. @mandarjog ?

cmd/istio-cni-daemonset/main.go
stewartbutler: >We already have a CNI daemonset, are users expected to run a second daemonset as well?

I could integrate this with the existing daemonset, by adding it to the install-cni.sh script.

cmd/istio-cni-daemonset/main.go
stewartbutler: I've changed both selectors to strings, and removed the supporting structs and code that I was using to generate the selector strings.

cmd/istio-cni-daemonset/main.go
stewartbutler: ICDNI changed to REPAIR, and repair code merged into existing daemonset as a new container.

deployments/kubernetes/Dockerfile.istio-cni-daemonset
stewartbutler: Using distroless.

deployments/kubernetes/cni-daemonset/istio-cni-daemonset.yaml
stewartbutler: Fixed the invalidity (when moving into other daemonset), but from what I can tell no alternate tags are being built. The CNI build pipeline doesn't appear to have run for a month or so. Going to leave it on `master-latest-daily` until I know what to change it to.

deployments/kubernetes/cni-daemonset/istio-cni-daemonset.yaml
stewartbutler: Merged with the existing daemonset in the newest version.

mandarjog: Adding a controller adds yet another component. Running cni-repairer as part of cni daemonset makes sense. It is solving a problem that was created by self.

stewartbutler: I removed it and merged it with the existing daemonset.

pkg/repair/repair.go
```diff
@@ -0,0 +1,275 @@
...
+		// If it is currently in a matching failed state , return true.
+		if state := container.State.Terminated; state != nil {
+			if state.Reason == "Completed" || state.ExitCode == 0 {
+				continue
+			} else if matchTerminationMessage(state) && matchExitCode(state) {
+				return true
+			}
+		}
+
+		// Next check that the container previously terminated due to our error
```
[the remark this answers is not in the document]
mandarjog: add comment: The LastTerminatedState  is necessary and sufficient for the way we are going to use it.

pkg/repair/repair.go
```diff
@@ -0,0 +1,275 @@
...
+// The pod reconciler struct. Contains state used to reconcile broken pods.
+type BrokenPodReconciler struct {
+	client  client.Interface
+	Filters *Filters
+	Options *Options
+}
+
+// Constructs a new BrokenPodReconciler struct.
+func NewBrokenPodReconciler(client client.Interface, filters *Filters, options *Options) (bpr BrokenPodReconciler) {
+	bpr = BrokenPodReconciler{
```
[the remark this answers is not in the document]
stewartbutler: Fixed

pkg/repair/repair.go
```diff
@@ -0,0 +1,275 @@
...
+
+	return
+}
+
+// Given a pod, returns 'true' if the pod is a match to the BrokenPodReconciler filter criteria.
+func (bpr BrokenPodReconciler) detectPod(pod v1.Pod) bool {
+	// Helper function; checks that a container's termination message matches filter
+	matchTerminationMessage := func(state *v1.ContainerStateTerminated) bool {
+		// If we are filtering on init container termination message and the termination message of 'state' does not match, exit
+		if s := strings.TrimSpace(bpr.Filters.InitContainerTerminationMessage); s == "" || s == strings.TrimSpace(state.Message) {
```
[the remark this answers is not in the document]
stewartbutler: Fixed

pkg/repair/repair.go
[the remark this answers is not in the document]
stewartbutler: I've updated the comment, if you want to take a look. I also streamlined the logic a bit.

deployments/kubernetes/install/helm/istio-cni/templates/istio-cni.yaml
```diff
@@ -164,3 +191,33 @@ subjects:
   name: istio-cni
   namespace: {{ .Release.Namespace }}
 {{- end }}
+    {{- if .Values.repair.enabled }}
```
howardjohn: nit: indent

howardjohn: how come we have the conditional here but either way we create the container?

deployments/kubernetes/install/helm/istio-cni/values.yaml
```diff
@@ -26,3 +26,15 @@ podAnnotations: {}
 # This can be used to bind a preexisting ClusterRole to the istio/cni ServiceAccount
 # e.g. if you use PodSecurityPolicies
 psp_cluster_role: ""
+
+repair:
+  enabled: true
+  hub: gcr.io/istio-release
+  tag: master-latest-daily
```
howardjohn: `latest`

deployments/kubernetes/install/helm/istio-cni/values.yaml
```diff
@@ -26,3 +26,15 @@ podAnnotations: {}
 # This can be used to bind a preexisting ClusterRole to the istio/cni ServiceAccount
 # e.g. if you use PodSecurityPolicies
 psp_cluster_role: ""
+
+repair:
+  enabled: true
+  hub: gcr.io/istio-release
```
howardjohn: gcr.io/istio-testing

howardjohn:
I mean lgtm aside from my comments. can be done in a follow up pr if easier

deployments/kubernetes/install/helm/istio-cni/templates/istio-cni.yaml
stewartbutler: Oversight, will fix it.

cmd/istio-cni-repair/main.go
```diff
@@ -0,0 +1,230 @@
...
+
+var (
+	loggingOptions = log.DefaultOptions()
+)
+
+// Parse command line options
+func parseFlags() (filters *repair.Filters, options *ControllerOptions) {
+	// Parse command line flags
+	// Filter Options
+	pflag.String("node-name", "", "The name of the node we are managing (will manage all nodes if unset)")
```
rlenglet: nit:
```suggestion
	pflag.String("node-name", "", "The name of the managed node (will manage all nodes if unset)")
```

cmd/istio-cni-repair/main.go
```diff
@@ -0,0 +1,230 @@
...
+	pflag.String(
+		"sidecar-annotation",
+		"sidecar.istio.io/status",
+		"An annotation key that indicates this pod contains an istio sidecar. All pods without this annotation will be ignored."+
+			"The value of the annotation is ignored.")
```
rlenglet: nit:
```suggestion
		"A pod annotation key that indicates that the pod has been injected with an Istio sidecar. All pods without this annotation will be ignored. "+
			"The value of the annotation is ignored.")
```

cmd/istio-cni-repair/main.go
```diff
@@ -0,0 +1,230 @@
...
+	pflag.String(
+		"init-container-name",
+		"istio-init",
+		"The name of the istio init container (will crash-loop if CNI is not configured for the pod)")
+	pflag.String(
+		"init-container-termination-message",
+		"",
```
rlenglet: Why does this flag have no default value?
If you don't know what a valid default value is, no one else will.
Just remove this flag.

cmd/istio-cni-repair/main.go
```diff
@@ -0,0 +1,230 @@
...
+	CreateEvents     bool            `json:"create_events"`
+	RunAsDaemon      bool            `json:"run_as_daemon"`
+}
+
+var (
+	loggingOptions = log.DefaultOptions()
+)
+
+// Parse command line options
+func parseFlags() (filters *repair.Filters, options *ControllerOptions) {
```
rlenglet: nit: Add the [Cobra command's `FlagSet`](https://godoc.org/github.com/spf13/cobra#Command.PersistentFlags) as a parameter and create the flags in there, for consistent with the rest of the Istio codebase.

cmd/istio-cni-repair/main.go
```diff
@@ -0,0 +1,230 @@
...
+	pflag.String(
+		"broken-pod-label-key",
+		"cni.istio.io/uninitialized",
+		"The key portion of the label which will be set by the reconciler if --label-pods is true")
+	pflag.String(
+		"broken-pod-label-value",
+		"true",
+		"The value portion of the label which will be set by the reconciler if --label-pods is true")
+
+	pflag.Bool("help", false, "Print usage information")
```
rlenglet: nit: Use [Cobra's built-in `help` command](https://github.com/spf13/cobra#help-command) like the rest of the Istio code, instead of implementing your own.

cmd/istio-cni-repair/main.go
```diff
@@ -0,0 +1,230 @@
...
+		DeletePods:       viper.GetBool("delete-pods"),
+		RunAsDaemon:      viper.GetBool("run-as-daemon"),
+		DaemonPollPeriod: viper.GetInt("daemon-poll-period"),
+		LabelPods:        viper.GetBool("label-pods"),
+		CreateEvents:     viper.GetBool("create-events"),
+		RepairOptions: &repair.Options{
+			PodLabelKey:   viper.GetString("broken-pod-label-key"),
+			PodLabelValue: viper.GetString("broken-pod-label-value"),
+		},
+	}
```
rlenglet: nit: Define constants for the flag names.

deployments/kubernetes/Dockerfile.istio-cni-repair
```diff
@@ -0,0 +1,10 @@
+# hadolint ignore=DL3007
```
rlenglet: You shouldn't need yet another image. We're trying to reduce the number of image.
Include this new binary into the image that is already being built.

deployments/kubernetes/install/helm/istio-cni/templates/istio-cni.yaml
```diff
@@ -132,6 +131,34 @@ spec:
               name: cni-bin-dir
             - mountPath: /host/etc/cni/net.d
               name: cni-net-dir
+      # This container installs the Istio CNI binaries
+      # and CNI network config file on each node.
+        - name: repair-cni
+          image: {{ .Values.repair.hub }}/istio-cni-repair:{{ .Values.repair.tag }}
```
rlenglet: Include the binary in the existing image and use it here:
```suggestion
          image: {{ .Values.hub }}/install-cni:{{ .Values.tag }}
```

deployments/kubernetes/install/helm/istio-cni/templates/istio-cni.yaml
```diff
@@ -132,6 +131,34 @@ spec:
               name: cni-bin-dir
             - mountPath: /host/etc/cni/net.d
               name: cni-net-dir
+      # This container installs the Istio CNI binaries
```
rlenglet: Fix this comment.

deployments/kubernetes/install/helm/istio-cni/templates/istio-cni.yaml
```diff
@@ -132,6 +131,34 @@ spec:
...
+            value: "true"
+          - name: "REPAIR_SIDECAR-ANNOTATION"
+            value: "sidecar.istio.io/status"
+          - name: "REPAIR_INIT-CONTAINER-NAME"
+            value: "istio-init"
+          - name: "REPAIR_BROKEN-POD-LABEL-KEY"
+            value: "{{.Values.repair.brokenPodLabelKey}}"
+          - name: "REPAIR_BROKEN-POD-LABEL-VALUE"
+            value: "{{.Values.repair.brokenPodLabelValue}}"
+
```
rlenglet: nit: Remove empty line.

deployments/kubernetes/install/helm/istio-cni/templates/istio-cni.yaml
```diff
@@ -164,3 +191,33 @@ subjects:
...
+kind: ClusterRole
+metadata:
+  name: istio-cni-repair-role
+  namespace: {{ .Release.Namespace}}
+  labels:
+    k8s-app: istio-cni-repair
+rules:
+- apiGroups: [""]
+  resources: ["pods", "events"]
+  verbs: ["get", "list", "watch", "delete", "patch", "update", "create" ]
```
rlenglet: Does this daemonset really need the permission to create pods in any namespace?
Trim this list to the minimum.
You should probably separate the `"pods"` and `"events"` permissions to have different verbs for each.

pkg/repair/repair.go
```diff
@@ -0,0 +1,273 @@
...
+				},
+				InvolvedObject: v1.ObjectReference{
+					Kind:            "pod",
+					Namespace:       pod.Namespace,
+					Name:            pod.Name,
+					UID:             pod.UID,
+					APIVersion:      pod.APIVersion,
+					ResourceVersion: pod.ResourceVersion,
+				},
+				Reason:         "BrokenIstioCNI",
```
rlenglet: I don't think "broken" is the right term. It's not informative to users.
I would suggest renaming this everywhere into "unitialized"

pkg/repair/repair.go
```diff
@@ -0,0 +1,273 @@
...
+				Message:        "Pod detected with broken Istio CNI configuration.",
```
rlenglet:
```suggestion
				Message:        "Initialization by Istio CNI failed.",
```

deployments/kubernetes/install/helm/istio-cni/templates/istio-cni.yaml
```diff
@@ -132,6 +131,34 @@ spec:
               name: cni-bin-dir
             - mountPath: /host/etc/cni/net.d
               name: cni-net-dir
+      # This container installs the Istio CNI binaries
+      # and CNI network config file on each node.
+        - name: repair-cni
```
rlenglet: As others commented, this should not run on every single node.
A Deployment with `priorityClassName: system-cluster-critical` and a single replica per cluster would be sufficient.

rlenglet:
My nit comments may be addressed in another PR.

cmd/istio-cni-repair/main.go
stewartbutler: This was to support the option of additionally specifying a termination message that could be written by the istio-init container by writing a message to `/dev/termination-log`. It is disabled by default (by leaving the default value as emptystring).

deployments/kubernetes/install/helm/istio-cni/templates/istio-cni.yaml
stewartbutler: I spoke with Romain offline and we resolved this, we believe that the fieldSelector on nodeName will be sufficient to address the performance concerns here.

deployments/kubernetes/install/helm/istio-cni/templates/istio-cni.yaml
stewartbutler: Split.

deployments/kubernetes/Dockerfile.istio-cni-repair
stewartbutler: Merged.

deployments/kubernetes/install/helm/istio-cni/templates/istio-cni.yaml
rlenglet: That's good. Thanks for the explanation.

haf:
Could you please make these design documents public by default? Thanks.

PR #231, opened 2020-01-22
Update dependencies to latest patch versions

dgn:


lambdai:
My guess github.com/containernetworking/cni v0.7.0 is the culprit of the test failure but I don't have any clue. Could you triage? @dgn

dgn:
> My guess github.com/containernetworking/cni v0.7.0 is the culprit of the test failure but I don't have any clue. Could you triage? @dgn

Yes it is. I can't update any of the dependencies without code changes I'm afraid; I also don't fully understand what we're testing in that failing test. Will close this for now.

PR #233, opened 2020-01-27
Release 1.5 codeowners

fpesce:


howardjohn:
Someone updated the build image used for the e2e test it looks like? or it depend s on istio/istio or something, but anyhow the kind creation looks incorrect

PR #237, opened 2020-01-30
Add ability to deploy config file standalone (not chained)

jwendell:
Addresses https://github.com/istio/istio/issues/20202


jwendell:
Istio / Operator side: https://github.com/istio/istio/pull/20708

PR #246, opened 2020-02-11
Watch-based repair controller

stewartbutler:
There are some performance issues in large clusters with the poll-based repair controller. This switches it to a watch-based controller. Creating the PR for feedback and so we have something to track.

Todo:
[ ] Add more tests
[ ] Possibly swap the controller from a daemonset to a singleton controller, depending on performance
[ ] Fix event generation (does not currently work in watch-based daemon mode)

### DIFF
--- a/pkg/repair/repair.go
+++ b/pkg/repair/repair.go
@@ -64,13 +64,10 @@ func NewBrokenPodReconciler(client client.Interface, filters *Filters, options *
 func (bpr BrokenPodReconciler) ReconcilePod(pod v1.Pod) (err error) {
 	log.Debugf("Reconciling pod %s", pod.Name)
 
-	if bpr.Options.LabelPods {
-		err = multierr.Append(err, bpr.labelBrokenPod(pod))
-
-	}
-
 	if bpr.Options.DeletePods {
 		err = multierr.Append(err, bpr.deleteBrokenPod(pod))
+	} else if bpr.Options.LabelPods {
+		err = multierr.Append(err, bpr.labelBrokenPod(pod))
 	}
 
 	return err
--- a/pkg/repair/repair.go
+++ b/pkg/repair/repair.go
@@ -31,6 +31,9 @@ const ValidationContainerName = "istio-validation"
 type Options struct {
 	PodLabelKey   string `json:"pod_label_key"`
 	PodLabelValue string `json:"pod_label_value"`
+	CreateEvents  bool   `json:"create_events"`
+	LabelPods     bool   `json:"label_pods"`
+	DeletePods    bool   `json:"delete_broken_pods"`
 }
 
 type Filters struct {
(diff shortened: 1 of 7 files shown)


pkg/repair/repair.go
```diff
@@ -59,32 +62,80 @@ func NewBrokenPodReconciler(client client.Interface, filters *Filters, options *
...
+	}
+
+	//if bpr.Options.CreateEvents {
+	//	newErr := bpr.createEventForPod(pod)
+	//	if newErr != nil {
+	//		errors = append(errors, newErr)
+	//	}
+	//}
+
+	errStr := strings.Join(errors, ",")
```
howardjohn: nit: ive seen `multierr` used before if you prefer and if its already imported here. no personal preference though

pkg/repair/repair.go
```diff
@@ -59,32 +62,80 @@ func NewBrokenPodReconciler(client client.Interface, filters *Filters, options *
...
 	}
 
 	for _, pod := range podList.Items {
-		labels := pod.GetLabels()
-		if _, ok := labels[bpr.Options.PodLabelKey]; ok {
-			log.Infof("Pod %s/%s already has label with key %s, skipping", pod.Namespace, pod.Name, bpr.Options.PodLabelKey)
-			continue
+		err := bpr.labelBrokenPod(pod)
+		if err != nil {
+			return err
```
howardjohn: should we continue on error?

pkg/repair/repair.go
```diff
@@ -59,32 +62,80 @@ func NewBrokenPodReconciler(client client.Interface, filters *Filters, options *
...
 		}
+	}
+	return nil
+}
 
-		if labels == nil {
-			labels = map[string]string{}
-		}
+func (bpr BrokenPodReconciler) labelBrokenPod(pod v1.Pod) (err error) {
+	if !bpr.detectPod(pod) {
```
howardjohn: is this redundant? if so maybe a comment would help

pkg/repair/repair.go
```diff
@@ -59,32 +62,80 @@ func NewBrokenPodReconciler(client client.Interface, filters *Filters, options *
...
+	if labels == nil {
+		labels = map[string]string{}
+	}
+
+	log.Infof("Labeling pod %s/%s with label %s=%s", pod.Namespace, pod.Name, bpr.Options.PodLabelKey, bpr.Options.PodLabelValue)
+	labels[bpr.Options.PodLabelKey] = bpr.Options.PodLabelValue
+	pod.SetLabels(labels)
+
+	if _, err = bpr.client.CoreV1().Pods(pod.Namespace).Update(&pod); err != nil {
+		log.Errorf("Failed to update pod: %s", err)
```
howardjohn: perhaps a personal preference, but I find the implicit return variable (not sure the proper name) a bit confusing, I thought we were ignoring there error here

pkg/repair/repair.go
```diff
@@ -179,6 +230,14 @@ func (bpr BrokenPodReconciler) DeleteBrokenPods() error {
 	return nil
 }
 
+func (bpr BrokenPodReconciler) deleteBrokenPod(pod v1.Pod) error {
+	if !bpr.detectPod(pod) {
+		return nil
+	}
+	log.Infof("Pod detected as broken, deleting: %s", pod.Name)
```
howardjohn: nit: include ns

pkg/repair/repaircontroller.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func (rc *RepairController) Run(threadiness int, stopCh <-chan struct{}) {
+	go rc.podController.Run(stopCh)
+	if !cache.WaitForCacheSync(stopCh, rc.podController.HasSynced) {
+		runtime.HandleError(fmt.Errorf("Timed out waiting for caches to sync"))
+		return
+	}
+
+	for i := 0; i < threadiness; i++ {
```
howardjohn: nit: seems unneeded since this is always 1, and we probably also always want it to be 1?

pkg/repair/repaircontroller.go
```diff
@@ -0,0 +1,124 @@
...
+		return false
+	}
+	defer rc.workQueue.Done(obj)
+
+	pod, ok := obj.(*v1.Pod)
+	if !ok {
+		log.Errorf("Error decoding object, invalid type")
+		if rc.workQueue.NumRequeues(obj) < 5 {
+			log.Infof("Requeueing failed object...")
+			rc.workQueue.AddRateLimited(obj)
```
howardjohn: this seems like the kind of error we don't want to requeue, and reason we should that I am missing? It will never succeed, will it?

pkg/repair/repaircontroller.go
```diff
@@ -0,0 +1,124 @@
...
+
+	if err == nil {
+		log.Debugf("Removing %s/%s from work queue", pod.Namespace, pod.Name)
+		rc.workQueue.Forget(obj)
+	} else if rc.workQueue.NumRequeues(obj) < 5 {
+		log.Errorf("Error: %s", err)
+		log.Infof("Re-adding %s/%s to work queue", pod.Namespace, pod.Name)
+		rc.workQueue.AddRateLimited(obj)
+	} else {
+		log.Infof("Requeue limit reached, removing %s/%s", pod.Namespace, pod.Name)
```
howardjohn: do we end up permanently broken here?

stewartbutler: Yep. I don't see a way to address it otherwise. If they are broken, we don't want them sticking around the work queue.

pkg/repair/repair.go
stewartbutler: Added

pkg/repair/repair.go
stewartbutler: Made more explicit

pkg/repair/repaircontroller.go
```diff
@@ -0,0 +1,119 @@
...
+func (rc *RepairController) Run(stopCh <-chan struct{}) {
+	go rc.podController.Run(stopCh)
+	if !cache.WaitForCacheSync(stopCh, rc.podController.HasSynced) {
+		runtime.HandleError(fmt.Errorf("Timed out waiting for caches to sync"))
+		return
+	}
+
+	go wait.Until(
+		func() {
+			for rc.processNextItem() {
```
howardjohn: what is the point of return code from processNextItem? seems like we always just loop forever regardless

stewartbutler: I'm not sure what you are asking -- if `rc.processNextItem()` returns `false`, the loop exits.

howardjohn: wait.Until will restart it though? you basically have `for{ for{ } }` is my understanding

stewartbutler: Oh, OK. This was based on https://github.com/kubernetes/client-go/blob/master/examples/workqueue/main.go, and that is how it works.

I'll remove the for loop.

stewartbutler: Actually, it doesn't appear to work, since wait.Until delays each run of the func by 1 second.

PR #252, opened 2020-02-17
Add ability to wait for config file to appear

srggavrilov:
### Problem
`istio-cni-node` pod may start up earlier than the main CNI provisioner (`aws-node` for example). In this case:
1. `YYY-istio-cni.conflist` will be used incorrectly for some time
2. `install-cni` container will terminate with an error and be restarted as the main CNI config appears

### Solution 1
If user explicitly provides `CNI_CONF_NAME` we could wait until  the latter has appeared. For that purpose another env variable has been introduced: `CNI_CONF_WAIT_SECONDS`

`CNI_CONF_WAIT_SECONDS` is equal 0 by default and doesn't affect default behaviour.

### Solution 2
Instead of falling back to default behaviour we could exit with error if `CNI_CONF_NAME` hasn't appeared in `CNI_CONF_WAIT_SECONDS` interval.

###  Implementation
This PR implements `Solution 1`.

With the following env:
```
      CNI_NET_DIR:            /etc/cni/net.d
      CNI_CONF_NAME:          10-aws.conflist
      CNI_CONF_WAIT_SECONDS:  30
```
you'd see something like:
```
Wrote Istio CNI binaries to /host/opt/cni/bin.
/host/secondary-bin-dir is non-writable, skipping
Using CNI config template from CNI_NETWORK_CONFIG environment variable.
CNI config: {
  "type": "istio-cni",
  "log_level": "info",
  "kubernetes": {
      "kubeconfig": "/etc/cni/net.d/ZZZ-istio-cni-kubeconfig",
      "cni_bin_dir": "/opt/cni/bin",
      "exclude_namespaces": [ "istio-system" ]
  }
}
10-aws.conflist doesn't exist; Waiting another 30 seconds.
10-aws.conflist doesn't exist; Waiting another 29 seconds.
10-aws.conflist doesn't exist; Waiting another 28 seconds.
10-aws.conflist doesn't exist; Waiting another 27 seconds.
10-aws.conflist doesn't exist; Waiting another 26 seconds.
10-aws.conflist doesn't exist; Waiting another 25 seconds.
10-aws.conflist doesn't exist; Waiting another 24 seconds.
10-aws.conflist doesn't exist; Waiting another 23 seconds.
10-aws.conflist doesn't exist; Waiting another 22 seconds.
10-aws.conflist doesn't exist; Waiting another 21 seconds.
10-aws.conflist doesn't exist; Waiting another 20 seconds.
Created CNI config 10-aws.conflist
Done configuring CNI.  Sleep=true
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,7 @@ Helm chart parameters:
 | cniBinDir | | `/opt/cni/bin` | Must be the same as the environment's `--cni-bin-dir` setting (`kubelet` param) |
 | cniConfDir | | `/etc/cni/net.d` | Must be the same as the environment's `--cni-conf-dir` setting (`kubelet` param) |
 | cniConfFileName | | None | Leave unset to auto-find the first file in the `cni-conf-dir` (as `kubelet` does).  Primarily used for testing `install-cni` plugin config.  If set, `install-cni` will inject the plugin config into this file in the `cni-conf-dir` |
+| cniConfWaitSeconds | | 0 | If set, `install-cni` will wait `cniConfWaitSeconds` seconds for the `cniConfFileName` to appear in the `cni-conf-dir` |
 | psp_cluster_role | | | A `ClusterRole` that sets the according use of [PodSecurityPolicy](https://kubernetes.io/docs/concepts/policy/pod-security-policy) for the `ServiceAccount`|
 | chained | `true` or `false` | `true` | Whether to deploy the config file as a plugin chain or as a standalone file in the conf dir. Some k8s flavors (e.g. OpenShift) do not support the chain approach, set to false if this is the case. |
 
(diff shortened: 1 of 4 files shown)


srggavrilov:
@googlebot I signed it!

srggavrilov:
@googlebot I fixed it.

srggavrilov:
@googlebot I fixed it.

srggavrilov:
Our case is a bit different. It's able to create the config but fails on https://github.com/istio/cni/blob/b9aa0b398c9f56586ac73f1ceb4c99401c4cdad0/deployments/kubernetes/install/scripts/install-cni.sh#L74
when the one from aws-node appears.

There's `CNI_CONF_NAME_OVERRIDE` which allows to skip this error but that is not what we want.

Perhaps there must be a loop starting from `find_cni_conf_file` and adding the istio block to whatever the current config file is.

PR #253, opened 2020-02-20
Removing redundant helm charts (istio/istio is source of truth now) 

stewartbutler:
Updating the CNI helm templates and the istioctl manifests (in a separate istio/istio PR) to make them consistent. Removing the event creation flag, since that functionality is removed.


stewartbutler:
It appears that the removal of the legacy helm charts from istio/istio has broken all the E2E tests in master. I'm recreating the PR direct to release-1.5.

@rlenglet -- should I repurpose this PR to just remove these helm charts altogether? Per our discussion offline, it sounds like these charts are superseded by the manifests in istio/istio now.

rlenglet:
> should I repurpose this PR to just remove these helm charts altogether? Per our discussion offline, it sounds like these charts are superseded by the manifests in istio/istio now.

@stewartbutler yeah, I'd remove the Helm charts altogether. You'll probably need to rewrite or delete `README.md` as well.

PR #257, opened 2020-02-26
#20044 Adding graceful exit trap to install-cni.sh

stewartbutler:
Fixes https://github.com/istio/istio/issues/20044

### DIFF
--- a/deployments/kubernetes/install/scripts/install-cni.sh
+++ b/deployments/kubernetes/install/scripts/install-cni.sh
@@ -30,6 +30,10 @@ exit_with_error(){
   exit 1
 }
 
+exit_graceful(){
+  exit 0
+}
+
 function rm_bin_files() {
   echo "Removing existing binaries"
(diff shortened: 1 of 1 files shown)


deployments/kubernetes/install/scripts/install-cni.sh
```diff
@@ -137,6 +141,8 @@ CFGCHECK_INTERVAL=${CFGCHECK_INTERVAL:-1}
 # Whether the Istio CNI plugin should be installed as a chained plugin (defaults to true) or as a standalone plugin (when false)
 CHAINED_CNI_PLUGIN=${CHAINED_CNI_PLUGIN:-true}
 
+trap exit_graceful SIGINT
```
rlenglet: Why not `cleanup`?

rlenglet: Nevermind, the `EXIT` trap will still execute.

PR #266, opened 2020-03-09
[WIP]: add annotation bindPodIPPorts support

ChenLingPeng:
issue: istio/istio#14738

using iptables rules to handle traffic from envoy to localhost, redirect it to using podIP.

The iptables rules will like this after applying this (`*` mark as new added line):
```
-A OUTPUT -p tcp -j ISTIO_OUTPUT
* -A ISTIO_BIND_PORT -j DNAT --to-destination 172.16.0.55
* -A ISTIO_BIND_PORT -j ACCEPT
* -A ISTIO_OUTPUT -d 127.0.0.1/32 -p tcp -m owner --uid-owner 1337 -m tcp --dport 6666 -j ISTIO_BIND_PORT
* -A ISTIO_OUTPUT -d 127.0.0.1/32 -p tcp -m owner --gid-owner 1337 -m tcp --dport 6666 -j ISTIO_BIND_PORT
* -A ISTIO_OUTPUT -d 127.0.0.1/32 -p tcp -m owner --uid-owner 1337 -m tcp --dport 7777 -j ISTIO_BIND_PORT
* -A ISTIO_OUTPUT -d 127.0.0.1/32 -p tcp -m owner --gid-owner 1337 -m tcp --dport 7777 -j ISTIO_BIND_PORT
-A ISTIO_OUTPUT -s 127.0.0.6/32 -o lo -j RETURN
-A ISTIO_OUTPUT ! -d 127.0.0.1/32 -o lo -j ISTIO_IN_REDIRECT
...
```

if this method is acceptable and istio/api#1314 is merge, `replace istio.io/api`will be removed and `[WIP]` will be removed

Signed-off-by: forrestchen <forrestchen@tencent.com>


ChenLingPeng:
An istio/istio PR will be sent if this is accepted.

howardjohn:
Why are we changing iptables rules instead of changing the clusters to point to pod IP instead of 127.0.0.1?

ChenLingPeng:
> Why are we changing iptables rules instead of changing the clusters to point to pod IP instead of 127.0.0.1?

@howardjohn there are 2 reason for me
1. the istio-iptables.sh exist in old version(1.4.* and before) istio/istio repo and the user could apply this changes very convinient to support this feature(1. pull proxyv2:1.4.5; 2. replace with new istio-iptables.sh into proxyv2 image; 3. edit inject template to handle new annotation)
2. istio using iptables(`-A ISTIO_OUTPUT ! -d 127.0.0.1/32 -o lo -j ISTIO_IN_REDIRECT`) to handle `pod to VIP itself`, I think the above rules is also needed even we changing cluster endpoint.

ChenLingPeng:
no need to change iptables rules if changing the clusters to point pod IP
/close

PR #287, opened 2020-04-02
align iptables args with non-cni

lambdai:
InboundPorts annotation set to star only if user doesn't provide label
Use user original value otherwise.

Signed-off-by: Yuchen Dai <silentdai@gmail.com>

Fix up according to #284

### DIFF
--- a/cmd/istio-cni/main.go
+++ b/cmd/istio-cni/main.go
@@ -177,10 +177,9 @@ func cmdAdd(args *skel.CmdArgs) error {
 			var containers []string
 			var initContainersMap map[string]struct{}
 			var annotations map[string]string
-			var ports []string
 			var k8sErr error
 			for attempt := 1; attempt <= podRetrievalMaxRetries; attempt++ {
-				containers, initContainersMap, _, annotations, ports, k8sErr = getKubePodInfo(client, string(k8sArgs.K8S_POD_NAME), string(k8sArgs.K8S_POD_NAMESPACE))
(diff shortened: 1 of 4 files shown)


cmd/istio-cni/main.go
```diff
@@ -224,7 +224,7 @@ func cmdAdd(args *skel.CmdArgs) error {
 				}
 				if !excludePod {
 					log.Infof("setting up redirect")
-					if redirect, redirErr := NewRedirect(ports, annotations); redirErr != nil {
```
rlenglet, line 227: `ports` is now unused.
Just remove the `ports` variable and modify `getKubePodInfo` to not return the ports. It should make `getK8sPodInfo` quite shorter.

cmd/istio-cni/redirect.go
```diff
@@ -202,10 +202,9 @@ func NewRedirect(ports []string, annotations map[string]string) (*Redirect, erro
 		log.Errorf("Annotation value error for redirect ports, using ContainerPorts=\"%s\": %v",
 			redir.includePorts, valErr)
 		return nil, valErr
-	} else if !isFound || redir.includePorts == "*" {
+	} else if !isFound {
```
rlenglet, line 205:
```suggestion
	}
	if !isFound {
```

cmd/istio-cni/main.go
lambdai: Great catch!

cmd/istio-cni/redirect.go
lambdai: Thanks!

PR #293, opened 2020-04-12
Add excludeIPRanges configuration

cagataygurturk:
Issue: istio/istio#19939

This PR supports configuring excludeIPRanges globally for the CNI plugin to bypass proxies.
 
Some Helm chart updates is required to inject the global.proxy.excludeIPRanges. I am gonna send another PR once these changes are incorporated.

includeIPCidrs, includePorts, excludeInboundPorts, excludeOutboundPorts are still missing.

### DIFF
--- a/cmd/istio-cni/main.go
+++ b/cmd/istio-cni/main.go
@@ -55,6 +55,7 @@ type Kubernetes struct {
 	NodeName             string   `json:"node_name"`
 	ExcludeNamespaces    []string `json:"exclude_namespaces"`
 	CniBinDir            string   `json:"cni_bin_dir"`
+	ExcludeIpRanges      string   `json:"exclude_ip_ranges"`
 }
 
 // PluginConf is whatever you expect your configuration json to be. This is whatever
@@ -222,7 +223,7 @@ func cmdAdd(args *skel.CmdArgs) error {
 				}
 				if !excludePod {
 					log.Infof("setting up redirect")
-					if redirect, redirErr := NewRedirect(annotations); redirErr != nil {
+					if redirect, redirErr := NewRedirect(annotations, conf); redirErr != nil {
 						log.Errorf("Pod redirect failed due to bad params: %v", redirErr)
 					} else {
 						log.Infof("Redirect local ports: %v", redirect.includePorts)
--- a/cmd/istio-cni/main.go
+++ b/cmd/istio-cni/main.go
@@ -55,7 +55,7 @@ type Kubernetes struct {
 	NodeName             string   `json:"node_name"`
 	ExcludeNamespaces    []string `json:"exclude_namespaces"`
 	CniBinDir            string   `json:"cni_bin_dir"`
-	ExcludeIpRanges      string   `json:"exclude_ip_ranges"`
+	ExcludeIPRanges      string   `json:"exclude_ip_ranges"`
 }
 
 // PluginConf is whatever you expect your configuration json to be. This is whatever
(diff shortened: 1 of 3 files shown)


rlenglet:
The Helm charts in this `istio/cni` repo are dead code in `master` / Istio 1.6. The only charts that are shipped in releases >1.5 are in https://github.com/istio/istio/tree/master/manifests/istio-cni. Please file a PR against that to be reviewed in parallel.

cagataygurturk:
@rlenglet a PR for charts is already there: istio/istio#22889

cmd/istio-cni/main.go
```diff
@@ -55,6 +55,7 @@ type Kubernetes struct {
 	NodeName             string   `json:"node_name"`
 	ExcludeNamespaces    []string `json:"exclude_namespaces"`
 	CniBinDir            string   `json:"cni_bin_dir"`
+	ExcludeIPRanges      string   `json:"exclude_ip_ranges"`
```
rlenglet, line 58: This option has nothing to do with this structure, which is specific to the Kubernetes environment.

rlenglet:
@cagataygurturk please confirm that adding a `traffic.sidecar.istio.io/excludeOutboundIPRanges` annotation into the `podRedirectAnnot` of your injection template works for you.

cagataygurturk:
@rlenglet Now I see that istio/istio#20472 is translating `global.proxy.excludeIPRanges` to podRedirectAnnot, which would indeed achieve what this PR trying to do. However something does not work for me.

Now I see that when I set `global.proxy.excludeIPRanges`, `podRedirectAnnot` is added to `istio-sidecar-injector` ConfigMap. However I noticed that `podRedirectAnnot` is not setting the annotations on the Pod after injection and 

What is more interesting, I see that `global.proxy.excludeIPRanges`'s value is correctly added in istio-validation init container (see below) and this container passes successfully. although in reality the IP block is not configured to bypass the sidecar proxy.

```
  - args:
    - istio-iptables
    - -p
    - "15001"
    - -z
    - "15006"
    - -u
    - "1337"
    - -m
    - REDIRECT
    - -i
    - '*'
    - -x
    - 188.166.192.215/32
    - -b
    - '*'
    - -d
    - 15090,15020
    - --run-validation
    - --skip-rule-apply
```

(188.166.192.215/32 is the CIDR I am testing with and I test all this for a cluster installed from the master branch.)

rlenglet:
> What is more interesting, I see that global.proxy.excludeIPRanges's value is correctly added in istio-validation init container to test the rules (see below) and this container passes successfully. although in reality the IP block is not configured to bypass the sidecar proxy.

When Istio CNI is enabled, the `istio-iptables` command is passed the `--run-validation --skip-rule-apply` options. In this case, all other options are ignored. That includes the `-x` options, which is also ignored.

I agree that this can be confusing. @lambdai it would be nice to remove all args except `--run-validation --skip-rule-apply` when Istio CNI is enabled, to avoid confusing users.

When Istio CNI is enabled, only the pod annotations matter.

@cagataygurturk If the annotations in `podRedirectAnnot` are not set on the pod after injection, please file a GitHub issue. That's a major bug.

cagataygurturk:
@rlenglet I am then closing this.

I have checked couple of times and in my cluster annotations are not injected. Filed an issue and I would love if another eye can see whether it is reproducible.

istio/istio#22902

cagataygurturk:
@rlenglet another user has just reported the same: istio/istio#22893

PR #299, opened 2020-04-17
Redirect: add 15090 to excludeInboundPorts 

lambdai:
As workaround of #23038 
Simulate what non-cni template injector does on port 15090.

excludeInboundPorts is read only if includeInboundPorts is '*' so we can blindly add 15090 regardless if excludeInboundPorts is used

Should revert this PR when `podRedirectAnnot` is propagated to webhook along with https://github.com/istio/istio/pull/23037

### DIFF
--- a/cmd/istio-cni/redirect.go
+++ b/cmd/istio-cni/redirect.go
@@ -226,11 +226,11 @@ func NewRedirect(annotations map[string]string) (*Redirect, error) {
 		return nil, valErr
 	}
 	// Add 15090 to sync with non-cni injection template
-	redir.excludeOutboundPorts = strings.TrimSpace(redir.excludeOutboundPorts)
-	if len(redir.excludeOutboundPorts) > 0 && redir.excludeOutboundPorts[len(redir.excludeOutboundPorts)-1] != ',' {
-		redir.excludeOutboundPorts = redir.excludeOutboundPorts + ","
+	redir.excludeInboundPorts = strings.TrimSpace(redir.excludeInboundPorts)
+	if len(redir.excludeInboundPorts) > 0 && redir.excludeInboundPorts[len(redir.excludeInboundPorts)-1] != ',' {
+		redir.excludeInboundPorts = redir.excludeInboundPorts + ","
 	}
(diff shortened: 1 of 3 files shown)


cmd/istio-cni/redirect.go
```diff
@@ -225,6 +225,12 @@ func NewRedirect(annotations map[string]string) (*Redirect, error) {
 			"excludeOutboundPorts", isFound, valErr)
 		return nil, valErr
 	}
+	// Add 15090 to sync with non-cni injection template
+	redir.excludeInboundPorts = strings.TrimSpace(redir.excludeInboundPorts)
+	if len(redir.excludeInboundPorts) > 0 && redir.excludeInboundPorts[len(redir.excludeInboundPorts)-1] != ',' {
+		redir.excludeInboundPorts = redir.excludeInboundPorts + ","
+	}
+	redir.excludeInboundPorts = redir.excludeInboundPorts + "15090"
```
howardjohn, line 233: Why add it here instead of https://github.com/istio/istio/blob/a671e62bd1ca3e6fd09bd0fbe03545c136b0f3f5/manifests/charts/istio-control/istio-discovery/files/injection-template.yaml#L406? This is where the other port, 15020, is defined?

stewartbutler: That may be my fault, I only saw the other annotation (which makes changes directly to the proxy sidecar) and assumed there was no similar way to pass that to the CNI pod. 

I can create a PR for the other, that would be a lower risk fix

lambdai: Oh, more ports?

lambdai: so we need another statusPort here in cni
and btw: where is `podRedirectAnnot` used?

lambdai:
Close in favor of https://github.com/istio/istio/pull/23037

stewartbutler:
Reopening, the other PR can't move forward due to another bug, so this has become the path of least risk

stewartbutler:
@howardjohn Can you give LGTM?

PR #319, opened 2020-05-15
More default exlusion ports

stewartbutler:
Adding more ports to the default set of excluded ports set in https://github.com/istio/cni/pull/299.

Also added a helper function to deduplicate ports, since I don't know what the behavior would be if we have ports declared multiple times.

### DIFF
--- a/cmd/istio-cni/main_test.go
+++ b/cmd/istio-cni/main_test.go
@@ -293,7 +293,7 @@ func TestCmdAddTwoContainersWithEmptyExcludeInboundPort(t *testing.T) {
 		t.Fatalf("expect using mockInterceptRuleMgr, actual %v", InterceptRuleMgrTypes["mock"])
 	}
 	r := mockIntercept.lastRedirect[len(mockIntercept.lastRedirect)-1]
-	if r.excludeInboundPorts != "15090" {
+	if r.excludeInboundPorts != "15020,15021,15090" {
 		t.Fatalf("expect excludeInboundPorts is \"15090\", actual %v", r.excludeInboundPorts)
 	}
 }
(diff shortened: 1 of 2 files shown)


stewartbutler:
Agreed. The problem here is that there is an issue in the way that istio/istio is passing global parameters in the meshconfig to CNI, and we've not had time to fix it. I'd estimate at least a few days to get a 'proper' fix in.

We have an open issue to fix it, but I lost track of it due to some competing priorities and didn't get it ready in time for 1.6. Once we have the proper fix written up in that issue, I'll create PRs cherrypick the fix to 1.4 thru 1.6 for their next patch releases.

mandarjog:
any reason for this to be draft?

stewartbutler:
I was waiting until the tests passed to swap it to non-draft, but I can do that now. My local setup got borked when my workstation crashed, so I'm having to lean on prow to run tests.

mandarjog:
Give me a cni image and I will verify it in my cluster .

mandarjog:
@stewartbutler how is `15020` already excluded ? Is there some one place where `15020` is defined?

stewartbutler:
@mandarjog 15020 was excluded indirectly by https://github.com/istio/cni/blob/master/cmd/istio-cni/redirect.go#L31. That default is added to excludeInboundPorts if no exclusions are set by the user.

I added 15020, 15021, and 15090 so that they are added to the list of excluded ports in all cases, even if the user defines a set of ports to exclude manually.

mandarjog:
ok, let's very fix before we build the whole thing..., give me a cni image, when ready. Thanks.

lambdai:
General LGTM
15020 is a preserved port for health check.
What is 15021?

lambdai:
NVM. I find 15021 in envoy bootstrap

PR #323, opened 2020-05-20
Reworking ipv6 detection logic

stewartbutler:
From `man hostname`:
```
-i, --ip-address
Display the IP address(es) of the host. Note that this works only if the host name can be resolved. Avoid using this option; use hostname --all-ip-addresses instead.
```
I've reworked the logic to use `--all-ip-addresses`, which shouldn't require a DNS query. Instead, we now loop across all IP addresses for the pod, and if we encounter one that is an ipv6 address, we enable ipv6 inbound rules.

This should fix https://github.com/istio/istio/issues/23770

### DIFF
--- a/tools/packaging/common/istio-iptables.sh
+++ b/tools/packaging/common/istio-iptables.sh
@@ -264,12 +264,14 @@ if [ -z "${PROXY_GID}" ]; then
 PROXY_GID=${PROXY_UID}
 fi
 
-POD_IP=$(hostname --ip-address)
-# Check if pod's ip is ipv4 or ipv6, in case of ipv6 set variable
+# Check if any of the pod's ip addresses are ipv6. If so, set variable
 # to program ip6tables
-if isIPv6 "$POD_IP"; then
-  ENABLE_INBOUND_IPV6=$POD_IP
-fi
+hostname --all-ip-addresses | while read -d' ' POD_IP; do
+  if isIPv6 "$POD_IP"; then
+    ENABLE_INBOUND_IPV6=$POD_IP
+    break
+  fi
+done
 
 #
 # Since OUTBOUND_IP_RANGES_EXCLUDE could carry ipv4 and ipv6 ranges
--- a/tools/packaging/common/istio-iptables.sh
+++ b/tools/packaging/common/istio-iptables.sh
@@ -266,10 +266,12 @@ fi
 
 # Check if any of the pod's ip addresses are ipv6. If so, set variable
 # to program ip6tables
-hostname --all-ip-addresses | while read -d' ' POD_IP; do
+ip -brief -oneline addr | awk '{print $3}' | sed 's|/.*||g' | while read POD_IP; do
   if isIPv6 "$POD_IP"; then
+    echo "Found ipv6 pod IP: ${POD_IP}"
     ENABLE_INBOUND_IPV6=$POD_IP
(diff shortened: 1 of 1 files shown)


lambdai:
```
       -I, --all-ip-addresses
              Display  all network addresses of the host. This option enumerates all configured addresses on all network interfaces. The loopback interface and IPv6 link-local addresses are omit‐
              ted. Contrary to option -i, this option does not depend on name resolution. Do not make any assumptions about the order of the output.
```
Looks like the dns is problem is resolved.

I have the concern since this is more aggressive on determining host as ipv6.
Statistically a cluster is more likely be ipv4. How about enabling ipv6 if there is no ipv4 address?

Of course the long term solution is to set up both ipv6 and ipv4 iptables

stewartbutler:
Thanks! I've still got some more testing I want to do before I swap it, but glad to have a sanity check.

stewartbutler:
OK, ended up having to switch from `hostname --all-ip-addresses` to using `ip addr` -- winds up that not all implementations of hostname support `--all-ip-addresses`. This changes the behavior somewhat, but is functionally equivalent.

I'm trying to figure out how to test with ipv6 on, I've never had an ipv6 cluster.

stewartbutler:
Glad I tested -- the loop I wrote in my first pass would've broken ipv6.

Winds up that if ipv6 is enabled on your machine, you can run kind in ipv6 mode to test. Seems to work fine now.

tools/packaging/common/istio-iptables.sh
```diff
@@ -264,12 +265,20 @@ if [ -z "${PROXY_GID}" ]; then
 PROXY_GID=${PROXY_UID}
 fi
 
-POD_IP=$(hostname --ip-address)
-# Check if pod's ip is ipv4 or ipv6, in case of ipv6 set variable
-# to program ip6tables
-if isIPv6 "$POD_IP"; then
-  ENABLE_INBOUND_IPV6=$POD_IP
-fi
+# Check if any of the pod's ip addresses are ipv6. If so, set variable
```
howardjohn, line 268: Wow I am surprised we are still using istio-iptables.sh here. Does that mean CNI has completely drifted from every iptables change made to Istio recently? Is there any efforts to use the golang implementation, and is this change needed there?

stewartbutler: I know that we use the go iptables implementation in the validation code, and we use it in the distroless image (when we don't have a bash to execute). 

I'm not sure if this change is needed in the go implementation.

I think we need to move cni into istio/istio, so we could work on migrating to the go binary when we make that change. That would support the effort to really get cni into wider use.

stewartbutler: (to clarify -- the distroless image with go binary passes some tests, but I don't know that they are comprehensive, and I'm not sure if anyone actually uses those in production)

stewartbutler: Actually, ignore that -- I'm confusing the install-cni container with the sidecar. I don't think istio-cni has ever been converted to use the go binary, but the sidecar image does use it for validating that the cni iptables rules have been applied correctly.

rlenglet: @howardjohn this has not drifted yet AFAIK. But that's definitely a high risk.

PR #339, opened 2020-06-03
add vendor dir to .gitignore

tariq1890:


tariq1890:
I use `go mod vendor` when developing locally. It would be nice to not add it

rlenglet:
> isn't vendor already gone?

Doh, you are right.

PR #348, opened 2020-12-09
Fixes BusyBox nsenter from parsing iptables script args

marshallford:
Prevents BusyBox's build of `nsenter` from incorrectly parsing `istio-iptables.sh` arguments as its own. Discovered when using the Kubernetes distribution "k3s" (https://github.com/k3s-io/k3s/issues/1434). A similar fix was used for [Linkerd](https://github.com/linkerd/linkerd2-proxy-init/pull/26).


howardjohn:
change seems reasonable but this repo is dead, thanks for reminding me we should archive it. Can you open in istio/istio (same code under "cni" repo)? Thanks!

marshallford:
Oops! Will do.

